Issue #73, opened 2009-08-13
Indexes and search should allow language specification


notanumber:
In order to support stemming in multiple languages, both `SearchBackend.update` and `SearchBackend.search` should allow an optional `language` parameter.

I've made a branch to explore this further.  It appears to be working, but I'd appreciate any input.

http://github.com/notanumber/django-haystack/tree/language

My current implementation extends `BaseSearchQuery` and `SearchIndex` to include an optional language (by default it will be the same as your current language as defined in settings.py).

sbaechler:
I created a backend for Elasticsearch which creates an index for every language defined in Django `settings.LANGUAGES`. Update and remove queries target all indexes, while the search query is routed to the index in the currently active language.
The analyzers for the text field are set to the specific language analyzers (if they are availabe) for building the index schema.
The multilingual backend is a drop-in replacement for the stock Elasticsearch backend.

https://github.com/sbaechler/django-multilingual-search

karolyi:
Might I suggest [my solution](https://github.com/notanumber/xapian-haystack/issues/174#issuecomment-339176959) based on a publicly available one?

Basically it's a multilingual wrapper indexing to separate databases for each languages. I updated the logic and added correct write lock handling (with `flock`) so multiprocess indexing could also work.

PR #394, opened 2011-07-27
Field collapsing / result grouping

albanm:
I added support for field collapsing with the solr backend.

cf http://groups.google.com/group/django-haystack/browse_thread/thread/7ad2466ce4f7993c?pli=1

SearchQuerySet has a collapse_by(self, field) function that will activate collapsing of results on the given field.

I didn't dive into haystack long enough to pretend it's a clean feature. But it works for me, it might deserve to be reviewed and amended if needed.

A few remarks for future work on this feature:

From what I have seen pysolr cannot run a query with grouping option activated without asking for a flat list of results (option group.main), the format returned without this option isn't supported and results aren't processed. Therefore the group value and size data is lost.

It would be nice also to make it easy to expand a group (e.g. 'more from this website'). It means running the same query after removing the collapse option and adding a filter on the current group value. I don't know if it deserves to be done by a template tag or directly in the search templates.

Alban

### DIFF
--- a/haystack/backends/__init__.py
+++ b/haystack/backends/__init__.py
@@ -6,6 +6,7 @@ from django.db.models import Q
 from django.db.models.base import ModelBase
 from django.utils import tree
 from django.utils.encoding import force_unicode
+from django.utils.translation import get_language
 from haystack.constants import DJANGO_CT, VALID_FILTERS, FILTER_SEPARATOR, DEFAULT_ALIAS
 from haystack.exceptions import MoreLikeThisError, FacetingError
 from haystack.models import SearchResult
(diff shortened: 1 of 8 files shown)


sbaechler:
@albanm I am giving the multilingual backend a new try. Your approach is similar to what I was looking for. I plan to do a backend for Elasticsearch.

acdha:
Closing this pull request as it's based against a very old version of the code. A current version would be interesting but we'd probably want to either support the analogous ElasticSearch feature or simply implement this by providing a way to get the native connection from the backend because the behaviour is not identical between the two search engines.

SalahAdDin:
what's about multilanguage support?

acdha:
@SalahAdDin That's a complicated problem which is really outside the scope of this issue. Since Haystack 2 added multiple backends you can use one core per-language (alluded to in http://django-haystack.readthedocs.org/en/latest/multiple_index.html#multi-lingual-content) but that'll still involve things like e.g. editing your Solr schema to configure all of the language-specific options appropriately for your content. 

#73 is the issue for having first-class support in haystack for specifying language settings at the index or field level.

PR #434, opened 2011-10-18
Lenient lookup: Added "must_exist" parameter to SearchField...

dannyadair:
Added "must_exist" parameter to SearchField. If set to False, lookup of a non-existent attribute will lead to an indexed value of None instead of raising an Exception.

I'm defaulting must_exist=True so haystack's old behaviour is preserved.

This would close https://github.com/toastdriven/django-haystack/issues/419

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -1,5 +1,6 @@
 from decimal import Decimal
 import re
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils import datetime_safe
 from django.template import loader, Context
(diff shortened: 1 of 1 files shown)


maikthomas:
Will this feature be integrated at any point? Is there a current workaround to simply deal with an object which is missing an attribute and give it a default?

acdha:
@maikthomas We really should have some tests before this could be merged. If this would be valuable to you, I would start by updating it against the current master branch and adding tests for both values of `must_exist`.

PR #466, opened 2011-12-23
Highlighting now works for exact searches (the ones with quotes)

kevinrenskers:
Added a simple replace('"', ''), so that highlighting now works when you search with quotes, like "hello world" instead of hello world.

### DIFF
--- a/haystack/utils/highlighting.py
+++ b/haystack/utils/highlighting.py
@@ -6,154 +6,154 @@ class Highlighter(object):
     html_tag = 'span'
     max_length = 200
     text_block = ''
-    
+
     def __init__(self, query, **kwargs):
         self.query = query
-        
+
         if 'max_length' in kwargs:
             self.max_length = int(kwargs['max_length'])
-        
+
         if 'html_tag' in kwargs:
             self.html_tag = kwargs['html_tag']
-        
+
         if 'css_class' in kwargs:
             self.css_class = kwargs['css_class']
-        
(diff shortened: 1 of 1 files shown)


SteveByerly:
thanks for the PR @kevinrenskers  i'm sorry you never received any feedback on this. 

i'll be trying to incorporate the exact query highlighting soon, but i think it should highlight the phrase exactly, instead of breaking up each word and highlighting separeately.

kopf:
Hey, I just happened to run into this problem too.

@SteveByerly: Any chance you could give me an approximate estimate as to how soon "soon" might be? :) If time is tight at the moment, and you have an idea of how this should be implemented, please just let me know, and I could take care of any heavy lifting for you. Thanks!

SteveByerly:
@kopf I'm hoping to get it merged into master by this weekend. If you need it sooner, it's actually really easy to override the default highlighter. Just create a new class that inherits `haystack.utils.highlighting.Highlighter` and point the `HAYSTACK_CUSTOM_HIGHLIGHTER` setting to its dotted path.

kopf:
@SteveByerly That's great news! If you remember, can you ping me a link to the PR once it's open - I'd just be curious to see what the implementation is like. I'm actually using a custom highlighter already, but would much rather hold out and have this fixed in haystack, rather than hacking a solution together myself on the application side.

SteveByerly:
I'll try to remember. The PR i'm working on is currently #1402

PR #498, opened 2012-02-03
Added as_view() method to SearchView, removed search_form_factory. Fixes #348.

### LINKED ISSUE #348: SearchView should have as_view() method
With Django 1.3 comes class-based generic views.  These all have an as_view() method that does the necessary thread-safe object creation.  Our as_view() could return a search_view_factory-like thing.

eppsilon:
I have added an as_view() method to the SearchView class, replacing search_form_factory.  The result is a slightly friendlier interface and some similarity with Django's class-based views.  This would fix issue #348.


SteveByerly:
I believe this can be closed since the inclusion of generic_views.py

acdha:
@SteveByerly Agreed – if there's anything else which should be added we should probably create separate issues by now

PR #509, opened 2012-02-14
Fixing rendering problem with highlighting overlapping terms, such as "what" and "a".

orblivion:
I found that searching for "what is a..." lead to the renderer to output "whatat" for what. It seemed to be tripping up the renderer that there were words within words. I thought at first to fix the renderer, but then I realized that it probably makes more sense to just make sure to feed it better highlight rules. Here's my thinking:

Get the earliest location of a matched term. If there are two starting within the same place (such as "hard" and "hardship") go with the longer one.

Potential problems with this are cases (which I can only come up with in theory) such as "anterritory" with "ant" and "territory" being search terms. Ideally, perhaps one would prefer "territory" be highlighted, since it's so much longer, but here "ant" will get highlighted. I thought about how to make "territory" highlighted, but I decided it was probably too complicated to be worth it, and not necessarily preferable anyway.

If, alternately, you want *everything* to be highlighted, (for instance, all of "anterritory" would be highlighted, despite it being a mashup of two terms, and not the term itself), you could go into the html rendering function and merge any of the overlapping highlight regions at the beginning of the function.

### DIFF
--- a/haystack/utils/highlighting.py
+++ b/haystack/utils/highlighting.py
@@ -30,28 +30,35 @@ class Highlighter(object):
     def find_highlightable_words(self):
         # Use a set so we only do this once per unique word.
         word_positions = {}
-        
+
         # Pre-compute the length.
         end_offset = len(self.text_block)
         lower_text_block = self.text_block.lower()
-        
-        for word in self.query_words:
-            if not word in word_positions:
-                word_positions[word] = []
-            
-            start_offset = 0
-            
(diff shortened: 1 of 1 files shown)


SteveByerly:
highlighting is currently being refactored in #1402 . I'm not sure how much of this extra work we want to do as I can see it creating performance issues.

If you want to submit another PR with tests and and documentation, we'll check it out again.

PR #540, opened 2012-03-26
Fixed UnicodeDecodeError for unicode PKs

syabro:
I have model with charfield as primary_key. With "bölüm" as value.
It produce
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)


claudep:
My feeling is that this sort of issues are no more with Python 3, feel free to reopen (with a test case) if it's still failing.

Issue #190, opened 2010-03-16
Highlighter find_window does not include words before the first hit


EmilStenstrom:
When looking at the default highlighter, I see that all highlighted snippets start at the highlighted word, and continue forward only. It would be more useful if the highlighter centered the word inside the search snippet, so you could understand more of the context.

It should be something like this (although this is not complete by far):

    from haystack.utils import Highlighter

    class CenteredWordHighlighter(Highlighter):
        def find_window(self, highlight_locations):
            best_start = 0
            best_end = self.max_length
        
            # First, make sure we have words.
            if not len(highlight_locations):
                return (best_start, best_end)
        
            words_found = []
        
            # Next, make sure we found any words at all.
            for word, offset_list in highlight_locations.items():
                if len(offset_list):
                    # Add all of the locations to the list.
                    words_found.extend(offset_list)
            
            if not len(words_found):
                return (best_start, best_end)
            
            return (max(0, words_found[0] - self.max_length/2), max(self.max_length,    words_found[0] + self.max_length/2))

sampeka:
It's been a long time and a ton of changes have been integrated since this issue was raised, however if you can reproduce this bug with haystack now please raise a new issue.

PR #549, opened 2012-04-09
Fix for issue #190: Highlighter find_window does not include words before the first hit

ciskeboekelo:
Hi Daniel,

It took me while to get around to it, but here's the code I wrote for issue #190.

You can see an example search result at:

http://loudestopinion.com/search/?q=wikipedia&x=0&y=0


sampeka:
Closing this PR due to the amount of time that has passed, lack of tests etc.

PR #565, opened 2012-05-03
A couple Solr specific improvements

akx:
These patches add support for an alternative schema template in build_solr_schema, and an option to change the Solr text field type (from the hard-coded default `text_en`) per-field.


akx:
Closing as positively ancient. 😆

PR #667, opened 2012-10-18
form_kwargs on basic_search

mativs:
Hi, in the basic_search method from the haystack/views.py file, the only customization that it's not avialable comparing with the SearchView class is the form_kwargs.

It's a very small change. I've just used the same code you used in the build_form method of the SearchView class.

Hope it helps

### DIFF
--- a/haystack/views.py
+++ b/haystack/views.py
@@ -177,7 +177,7 @@ class FacetedSearchView(SearchView):
         return extra
 
 
-def basic_search(request, template='search/search.html', load_all=True, form_class=ModelSearchForm, searchqueryset=None, context_class=RequestContext, extra_context=None, results_per_page=None):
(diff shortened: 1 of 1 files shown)


acdha:
@mativs I'm inclined to say this is moot given that 2.4 will the standard Django CBVs from #1130. Do you think it's worth updating this against master & adding some tests anyway?

mativs:
Uff this is very old, but I will try to take a look and add some tests.

acdha:
@mativs If it's not something you're interested in working on now, we can also just close it in favor of the new CBVs – I think it's a bit easier to do this with them.

mativs:
I will close it, I am not working with the project where I was using this change anymore and I would have to recreate everything again.

PR #705, opened 2012-12-20
Fix FacetedModelSearchForm

orf:
FacetedModelSearchForm inherits from ModelSearchForm, which means that the selected_facets kwarg is not popped anywhere down the chain until it gets passed to Form, which explodes because its not expecting it.

The fix is simple: make FacetedModelSearchFrom pop the "selected_facets" kwarg just like FacetedSearchForm.


SteveByerly:
Closing this in favor of #1400

PR #711, opened 2012-12-26
If only one found word in search, highight around

michaelmior:
It seems odd to me that when a single result is found, only text after
the term is highlighted. This change will center the found term in
search results.


michaelmior:
@bigjust Over six years later, here is the rebase and the tests!

PR #744, opened 2013-02-25
Allow for document boosting with elasticsearch

Connorhd:
Unless I'm mistaken document boosting doesn't work for elasticsearch without including this mapping.


Opalo:
@Connorhd is document boosting working with ElasticSearch? I've run into some serious problems when trying to use boosting with Solr. No I'm thinking about switching the search engine.

Connorhd:
@Opalo yep with this change document boosting worked great.

Issue #331, opened 2011-03-21
allow passing of language code to mangement commands/read it from settings


hvdklauw:
Currently the search indexes get generated in english, would be nice if we could pass the mangement command which translations should be loaded or if it's read from the settings.

That way you can generate a search index for a certain language

PR #745, opened 2013-02-26
Activate language in management commands

stefanw:
In order to be able to index stuff in a different language the translation must be activated explicitly. Otherwise "en-us" will be used and language in settings will be ignored.

This will not affect users who did not change Django's default LANGUAGE_CODE setting.

This would fix the problem in #331.


claudep:
Since https://github.com/django/django/commit/d65b0f72de8d35617fe0554ddabc950c7f323eef, Django does not force or deactivate translations, so it might have solved this issue.

PR #800, opened 2013-06-03
Delete index before update

juliomenendez:
On setup, if the index needs to be created and it already exists it raises

```
Failed to add documents to Elasticsearch: (400, u'IndexAlreadyExistsException[[index-name] Already exists]')
```


acdha:
This ticket is pretty old – can someone confirm whether this is still relevant with the current haystack & ES client library?

PR #816, opened 2013-07-01
Basic whoosh facet support

parruc:
Supports only base facets (no dates and query facets)

### DIFF
--- a/haystack/backends/whoosh_backend.py
+++ b/haystack/backends/whoosh_backend.py
@@ -318,8 +318,6 @@ class WhooshSearchBackend(BaseSearchBackend):
         if facets is not None:
             facets = [FieldFacet(facet, allow_overlap=True) for facet in facets]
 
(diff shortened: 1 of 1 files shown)


avoine:
We might want to disable stemming on faceted fields. It is enable by default for Text field in the whoosh backend see: 

https://github.com/django-haystack/django-haystack/blob/master/haystack/backends/whoosh_backend.py#L174

else:
Anyone willing to work on this?

PaLaMuNDeR:
I couldn't understand much if this feature is implemented already. I see that the first comment from 2013 is that it "Supports only base facets (no dates and query facets)", but when I try to use facets I get

/usr/local/lib/python2.7/site-packages/haystack/backends/__init__.py:33: Warning: Whoosh does not handle faceting.
  return func(obj, query_string, *args, **kwargs)
{}

I am using python 2.7.15, Django 1.11.15, whoosh 2.7.4, django-haystack 2.8.1
Could someone clarify how to make it work, please?

PR #826, opened 2013-07-13
Generic Views

lehins:
I've converted views to the django's generic views mixin style, which brings great advantages:
* views are now consistent with django style
* can be customized easily without completely overriding View classes
* mixin is possible with already existing Generic Views in other apps
* different Paginator class can be specified without even creating a new View class (by passing as a kwarg)
* completely backward compatible (maybe except unused method build_page and context_class property)

### DIFF
--- a/haystack/views.py
+++ b/haystack/views.py
@@ -1,5 +1,7 @@
+import warnings
 from django.conf import settings
 from django.core.paginator import Paginator, InvalidPage
+from django.views.generic.list import MultipleObjectMixin, ListView
 from django.http import Http404
 from django.shortcuts import render_to_response
 from django.template import RequestContext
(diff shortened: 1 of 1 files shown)


troygrosfield:
Did this PR die?  I came in search of a similar solution as I thought django-haystack's view's were quite odd compared to how django's current CBVs are being used.  Any updates on this?

troygrosfield:
It looks like another CBV implementation has already been done by @bennylope:

- https://github.com/bennylope/elasticstack/blob/master/elasticstack/views.py

@bennylope, why not submit a PR for the CBV to this repo since so many in the community are asking for it?

bennylope:
@troygrosfield the work I did for the elasticstack package didn't account for, and didn't need to account for, backwards compatibility. Any pull request would need to ensure that's not broken and tested as not broken. I don't have tremendous interest or time to do that right now.

For what it's worth I think a better solution would be to include Django CBV style mixins and views in a distinct module rather than swapping the classes. Then add deprecation warnings to the existing classes; point new users to the new classes; and eventually remove the old classes. If anyone has extended the existing class functionality then modifying these classes will break backwards compatibility without a lot of code yoga.

acdha:
+1 on @bennylope’s approach – if we have people who are really committed to the old views, I'd vote for moving them to a separate app to avoid anyone thinking this is recommended for code written after Django's own CBV implementation shipped.

Unfortunately my time to work on this is limited since I primarily use custom views but if someone wanted to contribute code I'd be happy to review it.

troygrosfield:
I pulled the CBVs from @bennylope's project into my project and with only minor changes, it works just as expected and was exactly what I was looking for.  I'm using django 1.7.2, python 3.4.2.

Looking at django-haystack's [.travis.yml](https://github.com/toastdriven/django-haystack/blob/master/.travis.yml) it looks like the minimum django version testing against is 1.5 which supports class based views.  

The other thought is to just make this change passive an include another file that contains the new CBVs.

- haystack
  - views.py <-- the current views left unchanged
  - class_based_views.py <-- includes the new CBVs

Then the change would be totally passive and new users would have to opt-in to the new approach.

acdha:
So perhaps the approach would be something like firing up the photocopier for @bennylope's views, adding the deprecation warnings to the old views, and updating the docs and tests. As part of the latter task it'd be good to review the options for customization to see if there are any obvious improvements to make.

it looks like unanimous agreement on adding a new module rather than shoe-horning it into the old one so we can do something like a major release cycle with the deprecation warnings before removing the old code. `class_based_views` seems like a fine name to me. Does anyone have a strong preference for something else?

lehins:
@acdha, how about `generic_views`, since old style views are also technically class based. Plus that is how django calls them as well.

troygrosfield:
@lehins, the only problem with ``generic_views`` is that they really aren't generic.  I guess you could argue they are generic views to search, but @bennylope's solution has concrete search views (i.e. ``FacetedSearchView`` and ``SearchView``) that, to me, doesn't fit the ``generic`` mold. I don't have a strong opinion either way as long as the view make it into django-haystack.

Thoughts?

lehins:
@troygrosfield, that's exactly how I was thinking, they are generic search views. But, it was just a suggestion. I hold the same point of view, most important they do make it to django-haystack.

troygrosfield:
@acdha, are you going to take the reins on this one?  Or would you like me to submit the PR for the new file?

acdha:
If you have time to work on this, I would appreciate it. My time available
is more at the level of reviewing it for the next couple months.

troygrosfield:
sounds good. I'll try and get the PR out shortly.

troygrosfield:
@acdha, @lehins, @bennylope PR has been submitted:

- https://github.com/toastdriven/django-haystack/pull/1130

Feel free to review.

SalahAdDin:
What is status of this?
Finish now?

troygrosfield:
It's finished waiting to be merged in #1130.

PR #851, opened 2013-08-23
UnifiedIndex.indexes is unpredictable for external callers

phill-tornroth:
I ran into a tricky bug whilst upgrading to Haystack2. My unit tests were failing because I have some stuff that depends on getting all of the indexes (something that used to be done via site.get_indexes()). When I looked into it I realized that ```connections[using].get_unified_index().indexes``` would be an empty dict if called early enough. After calling ```connections[using].get_indexed_models()``` (or anything that does the ```.build()``` indexes would be populated.

I think it might be worth considering something like what I'm scratched together here. Proposing a ```get_indexes()``` method for external (and even internal) callers that will do a build if needed. Along with renaming the ```index``` property to ```_index``` to make folks wary of using it.

Thoughts?


phill-tornroth:
@acdha This old thing :) I merged in master and resolved conflicts, in case you want to finger it for 2.4.0 or similar.

acdha:
Pending build with the updated commit: https://travis-ci.org/acdha/django-haystack/builds/49373420

PR #863, opened 2013-09-23
Upsert instead of overwrite during indexing

arski:
Hi there, this is not a complete patch but a proof of concept that works obviously. The idea is that a document in ES might be written to by several systems, and you don't necessarily want it to be fully replaced each time update_index is run from within Django.

Support for bulk_update on the pyelasticsearch side will hopefully be added soon via https://github.com/rhec/pyelasticsearch/pull/133

So this is basically something to consider, and hopefully agree that it is a good thing - then we can talk about making the PR more complete.

Cheers


arski:
im closing this, feel free to pick up if anyone wants

PR #916, opened 2013-12-10
Output proper class name in NotHandled message

claudep:


acdha:
This was fixed in 6c9024ca

PR #921, opened 2013-12-19
Use types in elasticsearch for differend haystack indexes.

### LINKED ISSUE #503: Use _type in ElasticSearch instead of django_ct
I don't know how flexible the architecture for this is, but it seems really... strange that we're not utilizing _type in ElasticSearch for each model.

This has the advantage of uploading a single mapping per model, easy commandline queries that are limited to a single model without using filters, etc.

Not sure how flexible the API is to handle this, but if it seems worthwhile, I could pursue it myself.

honzakral:
Fixes #503, thanks to Dave Dash for previous patch on which this one is
based.

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -118,7 +118,7 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
 
         unified_index = haystack.connections[self.connection_alias].get_unified_index()
         # Get mappings for all ElasticSearch types/models
-        for model, index in unified_index.indexes.iteritems():
+        for model, index in unified_index.indexes.items():
             es_type = self.get_es_type(index)
 
             self.content_field_name, field_mapping = self.build_schema(index.fields)


codingwithoutcomments:
I'm guessing this fix is dead in the water?

phill-tornroth:
@codingwithoutcomments Yeah, probably. It's a big change and Honza's focused on the Elasticsearch DSL library (which is worth checking out, last time I looked it was in good shape but missing the django-specific goodies).

Here is my custom backend which accomplishes the goals of this PR for my purposes. I don't use a lot of features though (faceting, more-like-this, etc.) so it's plausibly broken in a number of ways that just aren't affecting me. I've been running this backend for over a year though with relatively few changes. YMMV. Let me know if you have questions or find issues with it: https://gist.github.com/phill-tornroth/cd9a1a7c969c06eaae8b

phill-tornroth:
I should be more helpful. I think the most reasonable way to move the ES backend forward with this change (and other useful but pretty large changes) is probably not to massively edit the current ES backend in place, breaking everyone's indexes... but probably to take a backend like the one Honza started in this pull, or mine which I linked to - and introduce it as an _additional_ backend for Elasticsearch.

That way we could at least get the functionality out to folks and move forward without fear. Right now there's just not an ES focused maintainer actively on the project so I think it's a big leap of faith to merge changes like this one. It's also nasty for existing users because changes like this break existing indexes, which bums me out.

Anyways, that's my suggestion... Clean something like this up, copy/extend the existing ES tests and get it to pass a test suite, and introduce it as an additional backend available to elasticsearch. That lowers the risk to merge it and get it out there. If anyone is interested in putting that pull together I'd be happy to help it along by reviewing/testing, etc. Though to be clear I'm not a maintainer, just a fanboy of the original pull request.. so whether @acdha would merge a pull with that approach is up to him :)

davedash:
Hi, is this something we can close. It's 10 years old and I saw it in my "Pull Request" mentions :)

phill-tornroth:
Oh gosh yes please! It's super funny that I get notifications on it periodically too. I've long since migrated off of haystack.

claudep:
`Unsubscribe` button is your friend, I guess.

claudep:
I edited the initial description, maybe this will do the job?

PR #935, opened 2014-01-29
Make test Solr server work on systems without realpath

qris:


acdha:
Moot - start-solr-test-server no longer uses realpath

PR #959, opened 2014-02-27
Fix concurrency issues.

craigds:
We were getting this error a lot when under load in a multithreaded
wsgi environment:

    Model '%s' has more than one 'SearchIndex`` handling it.

Turns out the connections in haystack.connections and the
UnifiedIndex instance were stored globally. However there is a race
condition in UnifiedIndex.build() when multiple threads both build()
at once, resulting in the above error.

Best fix is to never share the same engine or UnifiedIndex across
multiple threads. This commit does that.

### DIFF
--- a/haystack/utils/loading.py
+++ b/haystack/utils/loading.py
@@ -4,8 +4,8 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import copy
 import inspect
+import threading
 import warnings
-
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.module_loading import module_has_submodule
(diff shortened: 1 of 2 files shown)


TheProdigyFilippo:
Hello! Anything progress in fixing this bug? I get the same error:
> Model xxx has more than one 'SearchIndex`` handling it
from time to time.

Regards

thedrow:
This needs to be rebased.
@TheProdigyFilippo If you need this fast you can take @craigds branch rebase and submit a new pull request.

craigds:
I've rebased against current master.

Thanks for the activity - I'd actually thought this issue was fixed already :)

acdha:
Has anyone looked into the test failures already?

https://travis-ci.org/django-haystack/django-haystack/jobs/109417309

acdha:
The test failures were due to one extra change where `_collections` was changed to `_local._connections` rather than `_local.connections` as used elsewhere. I corrected that.

I also renamed `_local` to `thread_local` in the hopes that this will avoid as much confusion as possible:

https://github.com/django-haystack/django-haystack/compare/master...acdha:4e1e2e1c5df1ed1c5432b9d26fcb9dc1abab71f4

thedrow:
:+1:

PR #960, opened 2014-02-27
Fix handling of slices in RelatedSearchQuerySet.

craigds:
We found whenever you do ?page=100 with RelatedSearchQuerySet, it will load all the results from all 100 pages in multiple solr queries (the exact number of solr queries is dependent on settings.HAYSTACK_ITERATOR_LOAD_PER_QUERY)

Needless to say that's not desired. This fixes the issues, and along the way gets rid of a heap of code duplicated from SearchQuerySet.


craigds:
closing in favour of #1269, which is updated and has passing tests

PR #971, opened 2014-03-11
Convert distance object to string for elasticsearch

petebrowne:
After upgrading to ElasticSearch 1.0, `geo_distance` searches broke. Apparently the default unit the geo queries changed to meters:

http://www.elasticsearch.org/guide/en/elasticsearch/reference/master/_parameters.html

To maintain backwards compatibility with 0.90.x we can convert the distance object to a string with an explicit unit.


petebrowne:
Digging through the source again, this is still not fixed? Should I make a new PR? What about merging in this one: https://github.com/toastdriven/django-haystack/pull/1003

This is a pretty critical bug...

acdha:
@petebrowne The code in #1003 was just merged after being updated against master & revising the tests + docs

PR #976, opened 2014-03-18
On update, check if object is include in index_queryset

foobacca:
and if not, then remove it.

This is useful when we are excluding some items, eg unpublished items.

This is the haystack 1.2.X version of #936.  Even if 1.2.X is no longer maintained, it means other people can find it if they need it.

### DIFF
--- a/haystack/indexes.py
+++ b/haystack/indexes.py
@@ -198,7 +198,12 @@ class SearchIndex(threading.local):
         """
         # Check to make sure we want to index this first.
         if self.should_update(instance, **kwargs):
-            self.backend.update(self, [instance])
+            # also check if this should be removed
+            qs = self.index_queryset()
+            if qs.filter(pk=instance.pk).count() > 0:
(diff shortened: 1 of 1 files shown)


mentholi:
Hello,

This seems to be problem even with the current version 2.6.0. I also assumed that `index_queryset()` would be used always. I realise that this pull request is for older version but very few changes it could work with current version also. 

Also there seems to be another issue open for this same thing. If the original maintainer doesn't have time to review these changes maybe you could get some help to get these fixes merged. Currently there is 965 forks of this project and I assume that is because fixes are not merged in to this version. This is not good for anyone!

Could you @foobacca update this fix to use newest master version? I could also create new pull request for this if there is some hope that it actually gets merged.

acdha:
If someone has time to add some tests, I'd appreciate it

mentholi:
@acdha I can do this when I have time to work with this again. Maybe next weekend.

PR #980, opened 2014-03-26
Add support for SQ in SearchQuerySet.narrow()

andreif:
Example:
```py
AND = lambda x: SQ(reduce(operator.__and__, x))
OR = lambda x: SQ(reduce(operator.__or__, x))

categories = ['foo', 'bar', 'hello', 'world']

q = OR([SQ(category=c) for c in categories])

sqs = SearchQuerySet().model(Product).narrow(q).facet('color')
```


acdha:
@andreif I had to make a few adjustments in https://github.com/acdha/django-haystack/commit/11ea90effc14164719a992fc88a6564291b64e27. The major one was changing tests to explicitly set the backend for each suite because the default backend wasn't configured under the new test environment and the base class raises NotImplementedError. 

While I was looking at that, I also tweaked your code slightly to use `_clone()` / `_reset()` to create the empty search query:

https://github.com/acdha/django-haystack/blob/11ea90effc14164719a992fc88a6564291b64e27/haystack/query.py#L415-L416

acdha:

Issue #704, opened 2012-12-20
__len__() should return >= 0 when using load_all()


gerdemb:
I am running Haystack 1.2.7 with xapian backend 1.2.12. Occassionally, I see the error "**len**() should return >= 0 " when using load_all() on a SearchQuerySet. Removing the call to load_all() solves the problem.  Example traceback:

```
>>> list(SearchQuerySet().models(models.Book).auto_query('harris').load_all())
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/gerdemb/lib/python2.7/haystack/query.py", line 128, in _manual_iter
    if self._cache_is_full():
  File "/home/gerdemb/lib/python2.7/haystack/query.py", line 99, in _cache_is_full
    if len(self) <= 0:
ValueError: __len__() should return >= 0
```

I see the following code in haystack/query.py 

```
    def _cache_is_full(self):
        if not self.query.has_run():
            return False

        if len(self) <= 0:
            return True
```

According to the Python data model documentation `__len__()` should return an an integer >= 0  so the code checking if the length is less than zero doesn't make any sense as Python will always raise a ValueError before the condition is tested.

http://docs.python.org/2/reference/datamodel.html#object.__len__

PR #999, opened 2014-05-09
'__len__' may never return a value < 0.

thomlinton:
There have been a few issues (#445, #704) and a referenced commit cd06d3be concerning the use of `__len__` in the `haystack.query` module.

In particular, the line:

`if len(self) <= 0`

will always raise a ValueError when `(self._result_count - self._ignored_result_count) < 0`. Although it appears that the referenced commit attempts to treat symptoms of this state, it seems that it is still quite possible that the implementation of `SearchQuerySet.__len__` will produce a negative value.

The attached pull request resolves this issue by ensuring that this method always returns a nonnegative integer.

Alternatively, the calling convention could be altered to avoid being beholden to the semantics bound to the `__len__` builtin:

`if self.__len__() <= 0`

Without more thoroughly reviewing the code paths, I'm unsure which method is a more practical resolution.

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -95,7 +95,7 @@ class SearchQuerySet(object):
                 self._result_count = 0
 
         # This needs to return the actual number of hits, not what's in the cache.
-        return self._result_count - self._ignored_result_count
+        return max([self._result_count - self._ignored_result_count, 0])
 
     def __iter__(self):
         if self._cache_is_full():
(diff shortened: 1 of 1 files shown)


frankystone:
Is someone working on this?

Using django-haystack 2.8.1 with Whoosh we are are getting server errors from time to time .. which we hadn't with haystack version 2.6.1.

Anything i can do?

acdha:
@frankystone The most recent pull request was #1601 which appears to work but didn't have any tests.

PR #1003, opened 2014-05-13
Support elasticsearch 1.x.x distance filter syntax

### LINKED ISSUE #957: ElasticSearch and distances
Hello,

Elasticsearch 1.0 only accepts distances as a string with unit integrated (for example "12km") while Django-Haystack uses a simple float.

Example Django-Haystack output : 

{"sort": [{"_geo_distance": {"location": [48.39, -4.4856], "unit": "km", "order": "asc"}}], "query": {"filtered": {"filter": {"bool": {"must": [{"terms": {"django_ct": ["utilisateurs.magasin"]}}, {"geo_distance": {"distance": 30.0, "location": {"lat": -4.4856, "lon": 48.39}}}]}}, "query": {"query_string": {"query": "(partenaire:(\"3\"))", "default_operator": "AND", "default_field": "text", "auto_generate_phrase_queries": true, "analyze_wildcard": true}}}}, "from": 0, "size": 1}

Origin of the Django-Haystack geo_distance :

https://github.com/toastdriven/django-haystack/blob/master/haystack/backends/elasticsearch_backend.py#L437

ElasticSearch reference for geo_distance :

http://www.elasticsearch.org/guide/en/elasticsearch/reference/master/query-dsl-geo-distance-filter.html#_options_3

### LINKED ISSUE #998: ElasticSearch >= 1.X support
I've wasted a lot of time figuring out why my Spatial queries stopped working after I've upgraded from ElasticSearch 0.90.X to version 1.X...
It seams django-haystack is not yet compatible with ElasticSearch 1.X, or am I mistaken?

dulaccc:
There were some *major* changes in the 1.0 release of Elasticsearch that break compatibility with haystack.
This PR is there to repair that :cake:

Cheers

Fix #957
Fix #998
Fix #984

### DIFF
--- a/docs/running_tests.rst
+++ b/docs/running_tests.rst
@@ -41,6 +41,13 @@ list of possible options::
 
 .. _nose: https://nose.readthedocs.org/en/latest/
 
+Or, to run the Elasticsearch backend's tests::
+
+    cd django-haystack/tests
+    export PYTHONPATH=`pwd`/..:`pwd`
+    export GEOS_LIBRARY_PATH=/opt/boxen/homebrew/lib/libgeos_c.dylib
+    export GDAL_LIBRARY_PATH=/opt/boxen/homebrew/lib/libgdal.dylib
+    django-admin.py test elasticsearch_tests --settings=elasticsearch_settings
 
 Configuring Solr
 ================
--- a/docs/running_tests.rst
+++ b/docs/running_tests.rst
@@ -41,14 +41,6 @@ list of possible options::
 
 .. _nose: https://nose.readthedocs.org/en/latest/
 
-Or, to run the Elasticsearch backend's tests::
-
-    cd django-haystack/tests
-    export PYTHONPATH=`pwd`/..:`pwd`
-    export GEOS_LIBRARY_PATH=/opt/boxen/homebrew/lib/libgeos_c.dylib
-    export GDAL_LIBRARY_PATH=/opt/boxen/homebrew/lib/libgdal.dylib
-    django-admin.py test elasticsearch_tests --settings=elasticsearch_settings
-
 Configuring Solr
 ================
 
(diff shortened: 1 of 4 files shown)


petebrowne:
Can this PR be merged in? I'm assuming this fixes the distance searching in ElasticSearch 1.x, right?

dulaccc:
Right @petebrowne, we use this patch on our server for about 8 months now and it works well, so for me we can merge it yes.
Besides, I've added a failing test as you had suggested in #971 and then fixed it.

docs/running_tests.rst
```diff
@@ -62,3 +61,11 @@ server is found all elasticsearch tests will be skipped. Note that the tests
 are destructive - during the teardown phase they will wipe the cluster clean so
 make sure you don't run them against an instance with data you wish to keep.
 
+If you want to run the geo-django tests you'll probably need some additional
+settings, just run those commands with your own lib paths::
+
+	cd django-haystack/tests
```
acdha: We need to update these paths for the new test_haystack path

docs/running_tests.rst
```diff
@@ -62,3 +61,11 @@ server is found all elasticsearch tests will be skipped. Note that the tests
...
+	export PYTHONPATH=`pwd`/..:`pwd`
```
acdha: This is only necessary in certain cases & should be omitted

docs/running_tests.rst
```diff
@@ -62,3 +61,11 @@ server is found all elasticsearch tests will be skipped. Note that the tests
...
+	export GEOS_LIBRARY_PATH=/opt/boxen/homebrew/lib/libgeos_c.dylib
```
acdha: This isn't necessary for most installations and is covered by the Django documentation:

https://docs.djangoproject.com/en/1.7/ref/contrib/gis/install/geolibs/#geos-library-path

I think it makes the most sense to remove this or at most put a pointer to the Django documentation rather than either distracting people or having to maintain a parallel copy of the Django docs.

acdha:
@dulaccc I think rebasing against master & cleaning up the tests a bit would be all this needs. It's a clean change except for the LIBRARY_PATH bits.

dulaccc:
Done. I've rebased on master and cleaned the documentation :)

acdha:

PR #1036, opened 2014-07-22
Fix code to enable highlighting

merwok:


SteveByerly:
I think the comments are clear you're supposed to add your own code. I think the PR makes it a bit more confusing.

merwok:
Would you accept the PR if I use ellipsis (`...`) instead of `pass`?  That would make the syntax valid and convey to readers that they need to fill in.

SteveByerly:
sure

merwok:
Changed!

PR #452, opened 2011-11-28
Allow custom analyzer for whoosh backend

adieu:
Currently the whoosh backend use StemmingAnalyzer for text fields. But sometimes we want to use other analyzer under some circumstances. For example, we may want to change the regular expression for the StemmingAnalyzer or we want a specific analyzer for some foreign languages.

This patch added analyzer argument for the SearchField. And user could use this argument to specify another analyzer other than the default one.

PR #1037, opened 2014-07-22
Analyzer function can be choosen for Whoosh backend

dex4er:
Haystack + Whoosh are not usable for inflectional languages. Standard StemmingAnalyzer is not enough but Whoosh allows to replace the analyzer very easly.

Unfortunately, Haystack hardcodes StemmingAnalyzer as an only analyzer. This patch allows to use own analyzer function.


denizdogan:
There is also #452 from 2011. Anyone looking at this?

PR #1056, opened 2014-08-15
Autocompleting numbers in Elasticsearch isn't working

phill-tornroth:
The "lowercase" tokenizer was being used for ngram/edge-ngram fields. Apparently it tokenizes numbers into individual digits, which are then ignored because the min-ngram size is 2. This means autocompletion of numbers just won't work at all. This commit contains a test which proves the bug and the fix. Essentially this is now tokenizing using "standard" and adds the lowercase operation as a filter.

Solution and understanding of the problem robbed from http://stackoverflow.com/questions/13636419/elasticsearch-edgengrams-and-numbers


acdha:
This looks good to me. I'm wondering whether there's any situation where lowercase wouldn't be appropriate but this matches our default in the Solr schema as well.

phill-tornroth:
I couldn't think of any issues, particularly since lowercase has already been used to tokenize.. and yeah, it matches Solr which is nice. I do think that this isn't a trivial change for people with existing ES environments to roll out though, since changing the tokenizer is equivalent to invalidating your existing indexes' data. Not sure what consideration that kind of thing gets with respect to major/minor releases, etc.

acdha:
Does it actually require you to rebuild the index to work at all, or only to get correct tokenization? My understanding was that if you install this and do nothing, you get the same behaviour as in the past for existing documents.

phill-tornroth:
Yes, that's my understanding as well. Good point, I guess it's not super
critical that you rebuild unless you need this fix.

PR #1090, opened 2014-10-21
Adds a --nocommit arg to the update_index, clear_index and rebuild_index management command.

andrewschoen:
This is helpful when doing bulk indexing and you'd rather have the backend handle commits, not the management command.  We had an instance where the amount of hard commits passed by update_index during a bulk index was causing stability issues with our solr cluster that was already doing soft and hard commits often.

I attempted to add this flag to clear_index and the --remove option of update_index, but it seems that solr was not respecting commit=false for those two code paths.  I can look into it some more, but wasn't able to find a solution.

### DIFF
--- a/docs/management_commands.rst
+++ b/docs/management_commands.rst
@@ -23,9 +23,17 @@ following arguments::
     ``--using``:
         If provided, determines which connection should be used. Default is
         ``default``.
+    ``--nocommit``:
+        If provided, it will pass commit=False to the backend.  This means that the
+        update will not become immediately visible and will depend on another explicit commit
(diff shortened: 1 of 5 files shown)


iElectric:
Another friendly ping. PR has docs and tests.

iElectric:
:beers:

viorels:
This now crashes for some backends, for example xapian, as they do not understand a "commit" argument for the update command. i think "commit" should be passed as an argument only to Solr (who makes sense of it) and not the other backends.

https://github.com/django-haystack/django-haystack/blame/master/haystack/management/commands/update_index.py#L91

PR #1093, opened 2014-10-23
Fix boolean values for Elasticsearch

joestump:
#### What's this PR do?

It makes `ElasticsearchSearchBackend._from_python()` treat boolean values the same way that `pysolr.Solr._from_python()` does.
#### Where should the reviewer start?

With the changes made to `ElasticsearchSearchBackend._from_python()` and the tests. It should be noted I changed one test in the ES test suite to match the Solr one.
#### How should this be manually tested?
1. Load up a shell with `./manage.py shell`
2. Create an `SearchQuerySet` instance and `filter` it by a `BooleanField` for a given model (e.g. `sqs = SearchQuerySet().models(MyModel).filter(my_boolean_field=False)`).
3. The output of `sqs.query.build_query()` should be `my_boolean_field:(false)` and not `my_boolean_field:(False)`.
#### Any background context you want to provide?

This discrepancy meant you had to effectively convert booleans to a string to use filtering on boolean values in Elasticsearch. I'm not sure if this was a conscious decision or not, but I did find it curious that the Solr backend behaved in opposition to how the ES backend did. I also believe this makes for a more Pythonic interface to instances of `SearchQuerySet` backed by the ES backend. 
#### What are the relevant tickets?

This supersedes #979.
#### What gif best describes this PR or how it makes you feel?

p.s. I modeled this PR on [a template we use at work I find useful](http://quickleft.com/blog/pull-request-templates-make-code-review-easier).

---

This change is [Reviewable](https://reviewable.io/reviews/django-haystack/django-haystack/1093)

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -664,6 +664,8 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
         elif isinstance(value, six.binary_type):
             # TODO: Be stricter.
             return six.text_type(value, errors='replace')
+        elif isinstance(value, bool):
+            return str(value).lower()
         elif isinstance(value, set):
             return list(value)
         return value
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -766,7 +766,7 @@ class ElasticsearchSearchQuery(BaseSearchQuery):
         prepared_value = value.prepare(self)
 
         if not isinstance(prepared_value, (set, list, tuple)):
-            # Then convert whatever we get back to what pysolr wants if needed.
+            # Then convert whatever we get back to what Elasticsearch wants if needed.
             prepared_value = self.backend._from_python(prepared_value)
 
         # 'content' is a special reserved word, much like 'pk' in
--- a/test_haystack/elasticsearch_tests/test_elasticsearch_backend.py
+++ b/test_haystack/elasticsearch_tests/test_elasticsearch_backend.py
@@ -1150,6 +1150,16 @@ class LiveElasticsearchRoundTripTestCase(TestCase):
         connections['elasticsearch']._index = self.old_ui
         super(LiveElasticsearchRoundTripTestCase, self).tearDown()
 
+    def test_boolean_is_converted_correctly(self):
+        """Ensure Python boolean values are correctly converted to ES string values."""
+        results = self.sqs.filter(is_active=True)
+        self.assertEqual(sqs.query.build_query(), u'is_active:(true)')
+
+    def test_string_booleans_still_work(self):
+        """Make sure the old way of converting to string first still works."""
+        results = self.sqs.filter(is_active='true')
+        self.assertEqual(sqs.query.build_query(), u'is_active:(true)')
+
     def test_round_trip(self):
         results = self.sqs.filter(id='core.mockmodel.1')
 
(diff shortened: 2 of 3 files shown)


acdha:
I confirmed that this merges cleanly and the tests pass in https://github.com/acdha/django-haystack/commit/56d1eb456cae1faa937872a2cadaca823c58e300. 

Next up is looking into the problem @phill-tornroth mentioned, starting with enhancing the round-trip tests.

acdha:
After looking at this further, I think converting all booleans to strings is going to lead to surprises elsewhere. In particular, I would like the value stored in the index to be of the expected type and value so that external queries still work using true booleans or 1/0 as in the ElasticSearch docs.

The code in #1141 (de896acdadc3010c7a96dd590ed871f68828548d) changes the behaviour slightly to pass the values through unquoted in `ElasticsearchSearchQuery.build_query_fragment`. All of the tests pass using this version and the mapping shows the correct type:

```json
"is_active": {
    "type": "boolean"
}, 
```

as we'd expect and the queries below all match the results using haystack:

```bash
http http://127.0.0.1:9200/test_default/modelresult/_search?q=is_active:true
http http://127.0.0.1:9200/test_default/modelresult/_search <<<'{"query": {"term": {"is_active": true}}}'
http http://127.0.0.1:9200/test_default/modelresult/_search?q=is_active:false
http http://127.0.0.1:9200/test_default/modelresult/_search <<<'{"query": {"term": {"is_active": false}}}'
```

I added a new test to avoid possible further regressions in the actual backend interface:

https://github.com/acdha/django-haystack/blob/de896acdadc3010c7a96dd590ed871f68828548d/test_haystack/elasticsearch_tests/test_elasticsearch_backend.py#L1201-L1224

I would appreciate someone else taking the time to review and test this. One of the questions I had is whether it's worth changing https://github.com/acdha/django-haystack/blob/de896acdadc3010c7a96dd590ed871f68828548d/haystack/backends/elasticsearch_backend.py#L757 to use the Raw input type so the value would be `true` instead of `(true)`. The catch here is that this would need to look up the field type in the index to know when to de-stringify `"true"` to have string queries match the bool types and it's not clear how this should work with e.g. the tests doing `content=False`. I'm inclined to say we only support boolean values on `BooleanField` out of general sanity and we could do something like add a DeprecationWarning if you make a string-value query on a boolean field. In any case, it works with the current `Clean` input type since ElasticSearch does the right thing if you have lower-case true and false.

phill-tornroth:
@acdha I quickly reviewed the change and tests. I'm not very familiar with `Clean()` so I'll need to go do some reading so that I understand why this works. However, the tests definitely look like they're proving the issue I saw so that's awesome. I also think it totally reasonable to only support boolean values on BooleanField.

I'll try to look a little deeper into the `Clean()` bit since this looks simpler than the solution I made on my fork, and will ping if I find anything fishy. Thanks for moving on this!

acdha:
I'm debating whether to hold up v2.4.0 for this. I think it's overdue but would like to get more testing / feedback on this. I've updated the code in #1141 against the current master & will look into adding some more tests later now that e.g. Travis is working even against Django 1.8.

joestump:
Closing due to inactivity — no updates in well over 60 days and no outstanding review engagement. Reopen if still wanted.

Executed via scheduled [Harness](https://github.com/stump-wtf/harness) using prism in [Crush](https://github.com/charmbracelet/crush)

PR #1098, opened 2014-10-28
Respect index_fieldname when converting values back to python from querying the backend

### LINKED ISSUE #891: SolrSearchBackend doesn't seem to honor the "index_fieldname" karg
[in `haystack/backends/solr_backend.py` in `SolrSearchBackend._process_results()`](https://github.com/toastdriven/django-haystack/blob/v2.1.0/haystack/backends/solr_backend.py#L380-L383):
1. this code:

``` python
  if string_key in index.fields and hasattr(index.fields[string_key], 'convert'):
    additional_fields[string_key] = index.fields[string_key].convert(value)
  else:
    additional_fields[string_key] = self.conn._to_python(value)
```
1. is executing in a context where:

``` python
  string_key="categories_descendant"
  value=['/News/Local']
  index=StoryIndex
```
1. and is breaking because: `string_key in index.fields` is `False`!
2. but my Index definition includes:
   `python
   categories = MultiValueField(index_fieldname='categories_descendant')
   `
3. section [1] should be rewritten as something like this

``` python
  fields = index.fields
  has_convert_fxn = lambda x: hasattr(x, 'convert')
  aliased = [x for x in fields if fields[x].index_fieldname == string_key]
  aliased = aliased[0] if aliased else False

  # Naive case where raw_data has a fieldname mentioned in the index
  if string_key in fields and has_convert_fxn(fields[string_key]):
    additional_fields[string_key] = fields[string_key].convert(value)

  # subtle case where raw_data has a fieldname which is aliased in the index
  elif aliased in fields and has_convert_fxn(fields[aliased]):
     additional_fields[string_key] = fields[aliased].convert(value)

  # default case
  else:
    additional_fields[string_key] = self.conn._to_python(value)
```

domenkozar:
Fixes #891 for all backends

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -587,8 +587,14 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
                     index = unified_index.get_index(model)
                     string_key = str(key)
 
-                    if string_key in index.fields and hasattr(index.fields[string_key], 'convert'):
-                        additional_fields[string_key] = index.fields[string_key].convert(value)
+                    # respect field.index_fieldname
+                    if string_key in unified_index.fields:
+                        instance_name = unified_index.fields[string_key].instance_name
+                    else:
+                        instance_name = string_key
+
+                    if instance_name in index.fields and hasattr(index.fields[instance_name], 'convert'):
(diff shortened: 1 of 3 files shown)


domenkozar:
Happy if someone picks this up and writes tests - we did have this deployed in production back in 2014 - but I no longer work for that company.

PR #1115, opened 2014-11-27
Make load_all() work with primary keys that are objects

linkmic:
Required to support django-uuidfield as primary key.


linkmic:
I made a small improvement. Still waiting on django-uuidfield to release with Python 3 fixes.

Issue #1097, opened 2014-10-27
LookupError with Django 1.7 and an app without models.py


deschler:
I'm in the process of upgrading a project to Django 1.7. Rebuilding or updating the haystack index fails in combination with django-grappelli's dashboard. Please note, that in latest grappelli, the dashboard module/app doesn't have a models.py (which is allowed in Django 1.7 as far as i understand).
 
    $ ./manage.py rebuild_index
    
    WARNING: This will irreparably remove EVERYTHING from your search index in connection 'default'.
    Your choices after this are to restore from backups or rebuild via the `rebuild_index` command.
    Are you sure you wish to continue? [y/N] y
    Removing all documents from your index because you said so.
    All documents removed.
    ERROR:root:Error updating grappelli.dashboard using default 
    Traceback (most recent call last):
    File "/path/to/env/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 189, in handle_label
        self.update_backend(label, using)
    File "/path/to/env/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 203, in update_backend
        for model in get_models(label):
    File "/path/to/env/local/lib/python2.7/site-packages/haystack/utils/app_loading.py", line 41, in get_models
        return [apps.get_app_config(app_label).get_model(model_name)]
    File "/path/to/env/local/lib/python2.7/site-packages/django/apps/config.py", line 166, in get_model
        "App '%s' doesn't have a '%s' model." % (self.label, model_name))
    LookupError: App 'grappelli' doesn't have a 'dashboard' model.
    Traceback (most recent call last):
    File "./manage.py", line 10, in <module>
        execute_from_command_line(sys.argv)
    File "/path/to/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
        utility.execute()
    File "/path/to/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 377, in execute
        self.fetch_command(subcommand).run_from_argv(self.argv)
    File "/path/to/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
        self.execute(*args, **options.__dict__)
    File "/path/to/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
        output = self.handle(*args, **options)
    File "/path/to/env/local/lib/python2.7/site-packages/haystack/management/commands/rebuild_index.py", line 16, in handle
[...]

PR #1124, opened 2014-12-15
Added extra model check to prevent that a simple module is treated as module

deschler:
Resolves an issue with Django 1.7 and django-grappelli. See #1097 for details.


acdha:
#1153 migrates away from having our own get_model implementation entirely

PR #1134, opened 2015-01-14
Handle missing database objects in {% highlight %}

acdha:
This avoids raising a non-obvious exception[1] when a search
result references a record which is no longer in the 
database. Instead the behavior will follow the default
template behavior for an invalid variable by returning
TEMPLATE_STRING_IF_INVALID.

1. See e.g. http://reinout.vanrees.org/weblog/2015/01/14/variabledoesnotexist-haystack-search.html or https://groups.google.com/forum/#!topic/django-haystack/0gjqYHb05u8


SalahAdDin:
Hi man, your lastest pull request and the troy... are a very good work, thank you very much.

reinout:
Yes, this would have fixed the issue I described in my blog entry.

What about a log message indicating the object doesn't exist? Otherwise it might be hard to track down the issue.

acdha:
@reinout You should already be getting that message from `_get_object` – see https://github.com/toastdriven/django-haystack/blob/74b9eba6fb572f6b3873a833535dfa42fc992795/haystack/models.py#L86. Is it possible that you don't have logging enabled for the `haystack` handler at logging.ERROR or are you hitting some alternate path which doesn't hit `_get_object`?

flevour:
@acdha I'm hitting this error when the object exists in the db but is ruled out by `index_queryset` (we have help articles with `published = False`).

acdha:
@flevour Is that with v2.3.1 or the current master branch? If it's still an issue, it'd be nice to get it fixed before 2.4 lands – if you have time to contribute a pull request or test that'd be great.

flevour:
@acdha running `1.2.7` on a 4-years old codebase (https://idonethis.com). I've applied the attached patch and site is running fine.

PR #1135, opened 2015-01-14
implementing _types, supporting python3, and currect database configuration

arsham:
Hello,

The latest three pull requests will break the tests, because base on my changes, now it can connect to ES server and tests fails. 
All my travis builds passed against python 2.6 to 3.4 and django 1.5 to 1.7

Cheers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: python
 python:
   - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
 
 cache:
   directories:
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -4,6 +4,7 @@ import datetime
 import re
 import warnings
 
+import django
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models.loading import get_model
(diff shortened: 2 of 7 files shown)


arsham:
Github reports it's failed, but please refering to [this](https://travis-ci.org/arsham/django-haystack) which is the built of my master branch, all (but pypy ones) passed.

SalahAdDin:
Perhaps @acdha can help you with it.

arsham:
@acdha hi mate, can you review the changes please?

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: python
 python:
   - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
```
acdha: These are redundant - we're already running the tests on Python 2.6-3.4 using tox – see the matrix in https://github.com/toastdriven/django-haystack/blob/74b9eba6fb572f6b3873a833535dfa42fc992795/.travis.yml#L29-L42

haystack/backends/elasticsearch_backend.py
```diff
@@ -731,6 +757,8 @@ def matching_all_fragment(self):
 
     def add_spatial(self, lat, lon, sfield, distance, filter='bbox'):
         """Adds spatial query parameters to search query"""
+        if six.PY3:
+            long = int
         kwargs = {
             'lat': lat,
             'long': long,
```
acdha: This change is actually unnecessary – this function has a bug because the variable being passed in is `lon` but the variable being used is the built-in `long` type!

.travis.yml
arsham: Fair enough

haystack/backends/elasticsearch_backend.py
acdha: I'm working on fixing the backend code now and updating the tests since we obviously didn't even enough coverage – ElasticSearch uses `lon` consistently and Solr appears to only use unnamed point pairs.

acdha:
Okay, the add_spatial code is now completely gone so the bug is moot:

https://github.com/acdha/django-haystack/commit/df8f5252a2e44f536919c070cc0a2274f06c9ac6
https://github.com/acdha/django-haystack/commit/621aac9e7be0f7ca8c99c1dac57ca8ef9909b15b

arsham:
Cheers mate, I will merge your branch to mine again, will edit based on your guidance, and will request a pull again.

PR #1137, opened 2015-01-20
Django master (and 1.8?) drops a number of relied upon methods/functions

kezabelle:
Django 1.8 drops `SortedDict` (in favour of `collections.OrderedDict`), `django.utils.importlib` and `django.db.models.loading.get_model`, all of which prevent haystack from loading.

I've not tested any of this, beyond getting `runserver` to complete and the root URLconf to load sucessfully by hitting a URL, because too many other packages I'm using also need fixing :/

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -6,7 +6,11 @@ import warnings
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models.loading import get_model
+try:
+    from django.db.models.loading import get_model
+except ImportError:  # Django > 1.8
+    from django.apps import apps
+    get_model = apps.get_model
 from django.utils import six
 
 import haystack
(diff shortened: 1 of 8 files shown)


acdha:
Hmmm, this will definitely require a major release since we'd have to drop Python 2.6 support (`collections.OrderedDict` was added in 2.7).

kezabelle:
A couple more commits let me get as far as viewing a list of SearchResults, highlighting text matches, and viewing all the fields in the SearchResult when checking using my [haystackbrowser app](https://github.com/kezabelle/django-haystackbrowser)

*NB: I changed the title to reflect the fact that in actuality I'm working against Django `master` rather than the `1.8.x` branches, so some of these moved/dropped imports may actually not stop-the-world until `1.9`, but I've not checked when the various affected bits are scheduled for deprecation.*

benjaminrigaud:
In 1.8:
> django.utils.translation.get_language() now returns None instead of LANGUAGE_CODE when translations are temporarily deactivated.

It breaks rebuild_index & update_index on my project.

acdha:
Haystack is now running against Django 1.8 and e.g. Python 2.6 support has been deprecated.  The locale issue should move to #745.

acdha:
This has been open for awhile — I'm going close this issue but please open new ones for anything which is still an issue in https://pypi.python.org/pypi/django-haystack/2.5.dev1.

PR #1141, opened 2015-01-30
ElasticSearch: avoid quoting bool values in queries

acdha:
This changes the way boolean fields are serialized for queries to generate
valid JSON true/false values rather than hitting ElasticSearch's implicit
conversion of all unrecognized values (i.e. not literal `0` or `false`) as
true.

See #1093


phill-tornroth:
Okay, now that I understand this more completely I can say with even more confidence that I'm a fan :)

acdha:
Thanks – I'd like to get this into 2.4.0 & get that shipped next week if possible so your time is greatly appreciated.

phill-tornroth:
@acdha I had written some tests in my own project to fix my own fix for this bug (my fix was nastier... I introduced an argument to `_from_python` for whether we were converting values for query or document and then returned "true"/"false" for queries only). At any rate, I tested your fix and most tests pass but one set of cases failed. I think they came up in someone's issue report which inspired me to check them. They're admittedly pretty edge-casey:

`SearchQuerySet(MyModel).filter(verified__in=[False])`

Doing an `__in` clause seems to fail in your case. I don't fully remember what inspired me to even write that test, like I said I think it was a discussion I found in another issue or pull request discussing possible fixes for this and their shortcomings. I'm also not sure if knowing that this test fails might help you guess some more realistic problems the fix has... so wanted to mention it.

phill-tornroth:
Ah, it looks like this is pretty easily resolved. We just need to repeat the check for boolean values lower in build_query_fragment where we deal with `__in` clauses. I also don't know if we need to handle `__exact`. Both seem a bit silly, but I'm not sure if there aren't some esoteric patterns folks might want to use dynamically across fields of different types that could stumble on this.

But yeah, I think something like this ought to do it:

```python
            elif filter_type == 'in':
                in_options = []

                for possible_value in prepared_value:
                    if isinstance(possible_value, bool):
                        possible_value = Clean(str(value).lower())
                    in_options.append(u'"%s"' % self.backend._from_python(possible_value))

```

phill-tornroth:
Err...This begs the question whether we have a sanitized string issue with `__in` clauses also, doesn't it? Since we're not `Clean()`-ing string values in this path, unless I'm misunderstanding something.

PR #1149, opened 2015-03-03
FacetedSearchMixin bugfixes and improvements - facet_fields attribute added

lovmat:


SalahAdDin:
But, extra content was deprecated, not is get_queryset.

narrowfail:
I need this guys. 

Thanks!

MichaelAnckaert:
Can we get the pull request referenced above merged please? The generic view FacetedSearchView is essentially useless at this point.

SalahAdDin:
Why @MichaelAnckaert ?

acdha:
The only failing build is for Django 1.5, which was removed from our support matrix in https://github.com/django-haystack/django-haystack/commit/8b368ff2ed4190926c7db0e04c5e473cea9fcfda.

haystack/generic_views.py
```diff
@@ -88,9 +88,10 @@ class FacetedSearchMixin(SearchMixin):
     faceting.
     """
     form_class = FacetedSearchForm
+    facet_fields = None
```
sframe: Thanks for the patch. I used this update within my project as a work around for the bug.

I also liked the idea of adding facet_fields attribute to view. For my project I needed to pass additional args for sorting and limits.

How about allowing this to be an tuple of dicts, then unpack each dict for the args in the facet method?

```python
facet_fields = ({'field': 'author', 'sort': 'index', 'limit': -1},)
```

```python
def get_queryset(self):
    qs = super(FacetedSearchMixin, self).get_queryset()
    for field in self.facet_fields:
        qs = qs.facet(**field)
    return qs
```

PR #1165, opened 2015-04-06
Resolve various Django 1.8 deprecations

benspaulding:
This pull request handles the deprecation in Django 1.8 of:

- `django.utils.datastructures.SortedDict`
- `django.utils.importlib`
- `django.utils.unittest`

For each case it prefers code from the Python stdlib, and if that import fails uses the old Django code. You can see individual commit messages for more information on each change.

This should resolve issue #1162 as well as pull request #1137.

PR #1175, opened 2015-04-25
Updating haystack to test against django 1.8

troygrosfield:
Get django-haystack to run tests against django 1.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ env:
     - TOX_ENV=py34-django1.5
     - TOX_ENV=py34-django1.6
     - TOX_ENV=py34-django1.7
+    - TOX_ENV=py34-django1.8
     - TOX_ENV=pypy-django1.5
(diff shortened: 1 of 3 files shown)


troygrosfield:
The pull request fails because django 1.8 isn't supported yet.  There are other PRs open that addresses a lot of these issues [1].  However, it's important to see how close this lib is to having support for django 1.8 so I would recommend still merging this.  That way we can see how future pull requests are doing against django 1.8.

If nothing else, then we could add ``py34-django1.8`` build as an allowed failure for now until all issues are resolved.

@acdha, thoughts?

[1] https://github.com/django-haystack/django-haystack/pull/1165, https://github.com/django-haystack/django-haystack/pull/1137

acdha:
The Django 1.8 branch has merged as of bd8e512db0d6703ecaa95effcabb460f5400c79d

acdha:

PR #1177, opened 2015-04-28
Update TravisCI link on README

paulshannon:
I think the repo got changed at some point and the old project referenced at travisci doesn't exist anymore...


acdha:
Ah, yes, that'd be when the project moved from Daniel's personal account to the django-haystack organization.

Issue #1167, opened 2015-04-08
django.db.close_connection() removed in Django 1.8


a5huynh:
As noted in the [release notes here](https://docs.djangoproject.com/en/1.8/releases/1.8/).

Using `Django==1.8` and `django-haystack==2.3.1` this unfortunately messes up a `rebuild_index/update_index` with `-k/--workers` set to anything larger than 0.

Example traceback:

```
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/opt/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 330, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/base.py", line 390, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/base.py", line 441, in execute
    output = self.handle(*args, **options)
  File "/opt/env/lib/python2.7/site-packages/haystack/management/commands/rebuild_index.py", line 16, in handle
    call_command('update_index', **options)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 120, in call_command
    return command.execute(*args, **defaults)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/base.py", line 441, in execute
    output = self.handle(*args, **options)
  File "/opt/env/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 184, in handle
    return super(Command, self).handle(*items, **options)
  File "/opt/env/lib/python2.7/site-packages/django/core/management/base.py", line 619, in handle
    label_output = self.handle_label(label, **options)
  File "/opt/env/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 189, in handle_label
    self.update_backend(label, using)
  File "/opt/env/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 215, in update_backend
    db.close_connection()
AttributeError: 'module' object has no attribute 'close_connection'
```

acdha:
This should be fixed in https://github.com/acdha/django-haystack/tree/django-1.8 which has the patch from #1184.

PR #1184, opened 2015-05-04
fix 1.8 incompatibility #1167

phingage:
fix 1.8 incompatibility with deleted db.close_connection() internal API. see #1167

### DIFF
--- a/haystack/management/commands/update_index.py
+++ b/haystack/management/commands/update_index.py
@@ -8,7 +8,7 @@ import warnings
 from datetime import timedelta
 from optparse import make_option
 
-from django.db import connection
+from django import db
 from django.core.management.base import LabelCommand
 from django.db import reset_queries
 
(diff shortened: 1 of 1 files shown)


acdha:
This can't be merged in its current state because it's not compatible with Django 1.7 and earlier.

acdha:
I'd like to drop support for older versions of Django in the not terribly distant future – maybe just toss in a deprecation warning when we have to fall back to `close_connection` and we can drop that in a future major release?

phingage:
@acdha  71d6125 should be compatible with older version of django.

phingage:
@acdha Should be ok now (405bbf8)!

acdha:
The Django 1.8 branch has merged as of bd8e512db0d6703ecaa95effcabb460f5400c79d

acdha:

PR #1189, opened 2015-05-11
Documentation update: Add get_results note

yigitguler:
I spend a good amount of time while trying this method work. After seeing that i sholud actually call get_results to limit the results, i wanted to share this information with other users. Thanks!

### DIFF
--- a/docs/searchquery_api.rst
+++ b/docs/searchquery_api.rst
@@ -215,6 +215,9 @@ by chaining this method several times.
 
 Restricts the query by altering either the start, end or both offsets.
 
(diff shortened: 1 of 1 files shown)


acdha:
I think this needs more context. What behaviour were you seeing? If you're calling `set_limits` on a `SearchQuery` which is associated with at `SearchQuerySet` before the query has run the values will be reset to the current slice (see https://github.com/django-haystack/django-haystack/blob/master/haystack/query.py#L170). Calling `get_results()` changes that because a query instance won't run more than once but could cause confusion because *all* subsequent changes to the queryset won't be reflected.

Issue #1194, opened 2015-05-18
Race condition when running update_index --remove


diwu1989:
Check out this line:
https://github.com/django-haystack/django-haystack/blob/master/haystack/management/commands/update_index.py#L281

You are paginating over the search queryset from 1 to N, and at the same time, you're removing stale documents. Let's say that you're on page X, and you removed N documents, when you move forward to the next page X+1, you would have missed N documents that would have been on the next page, but they moved forward because your performed the deletion from the previous page.

I think you should do a pass through to find the document ids that should be removed, and then perform the actual remove afterwards in a separate pass.

acdha:
The code you referenced is old enough that memory usage was a more pressing concern, particularly since some haystack users have things like UUIDs instead of integer keys. The two-pass approach is probably sufficient but I'm wondering whether some users regularly delete enough records that it'd be better to reset the step to start at the first deleted record after we process all of the removes for a block so we never have more than one block in memory at a time.

One other note which comes to mind is that I don't see enough benefit to using workers for removes – unlike the indexing, there's no substantial overhead on our side for deleting a record and it'd simplify the code not to have the extra path.

The code which I use in production loads everything into sets so it can efficiently determine all of the pks which need to be added or removed from the index, which is fine since the number of records & size of keys is well within the server memory budget:

https://gist.github.com/acdha/08d0a760221184bd84cf#file-sync_index-py-L82-L83

diwu1989:
Agree that using multiple worker is not worth the extra effort.

If you want to save memory, you can do:
- grab a page of documents from search index order_by document identifier
- ask ORM for `.filter(pk__in=pk_list).values_list('pk', flat=True)` and then find set difference
- take the greatest document identifier of the working set
- repeat with the next page of documents that has identifier greater than the max seen, stop when page is empty

pa-jama:
Whenever I run update_index --remove, it always hangs at the end, no matter how many workers, is that related to this issue?

acdha:
@pa-jama If you're using workers > 1 on old versions of python, potentially. Otherwise it's hard to say without a testcase

acdha:
@diwu1989 I took the simple approach for https://github.com/django-haystack/django-haystack/commit/3c88d41f9a02299d031eef8114dfd4fdc5ed358f because the incremental logic was complicated by the need to deal with running with `commit=False` and I wasn't happy with the amount of complexity that version added to save a modest amount of memory.

PR #1201, opened 2015-05-26
Fix issue with "MySQL has gone away error"

ebressler:
This fixes an issue with "MySQL has gone away error" when using -r with -k when doing an update Index.

### DIFF
--- a/haystack/management/commands/update_index.py
+++ b/haystack/management/commands/update_index.py
@@ -256,6 +256,11 @@ class Command(LabelCommand):
                 pool.join()
 
             if self.remove:
+                # For some reason, if we don't close them all out, we find out that the connection
+                # has gone away. I assume this is because the remove is happening in the root
+                # process and that the connections are messed up.  This only happens when using 
+                # the -k or --workers flag.
+                db.close_connection()
                 if self.start_date or self.end_date or total <= 0:
                     # They're using a reduced set, which may not incorporate
                     # all pks. Rebuild the list with everything.


ebressler:
As far as I can tell, the CI failures are not a result of this PR but of existing failures before it.

acdha:
Yeah, the Travis CI hosts go through periods of being flaky but if you run the tests locally, everything works.

Are you using persistent connections? Remove is currently using workers so I suspect what we're seeing is related to passing state over from the master process. Long term it makes the most sense to simply not use workers for removal since, unlike updates, there's no significant work to be done in the removal thread (see #1194).

ebressler:
Yes, I am using workers and this problem only happens when I run with both workers and remove. I do persist connections and I tried all the setting to make it not persist and that did not seem to fix the issue.  

I see now that basically a second round of workers is made and this causes the problem I believe as these are new sets of workers and a query happened in the parent process before going back into the ghetto queue and forking again.

Anyway, this does the same thing that was done multiple lines ahead of it to make sure that the connections are good and since removal is really only 1 more query to the DB as far as I can tell, it is not a big deal to close them off to avoid the persistence problem.

Anyway, anything I need to do to make this accepted into the main tree so I don't have to run off a fork or is there another preferred solution?

ebressler:
Can I get some feedback on what is necessary to get this into master @acdha?

acdha:
Just time on my part

ebressler:
@acdha Thanks.  I am looking forward to it being on mainline so we don't have to work on a fork.

PR #1206, opened 2015-06-06
Fixed get_model and models.loading deprecation warnings for Django 1.8

Koed00:
- Backends were still using `get_model` from `django.db.models.loading`. Since Django 1.7 this has been deprecated in favor of the new `apps.get_model`. The old version will be deprecated in Django 1.9 and currently gives deprecation warnings. This replaces that with a shadow function that returns the correct function based on the django version used.

- `haystack_get_models` was still using the `django.db.models.loading.get_model` in the Django > 1.7 explicit block, thus creating warnings.
I've replaced this with code that only uses the apps variant. 

- Imports for Django < 1.7 were giving deprecation warnings in Django > 1.7. Imports have been moved down to the appropriate block

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -8,7 +8,6 @@ import warnings
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models.loading import get_model
 from django.utils import six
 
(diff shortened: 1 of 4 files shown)


acdha:
Thanks – this merged & 2.4.0rc2 is now on PyPI

PR #1207, opened 2015-06-08
Bug fix: ensure raw_result won't raise a KeyError for the _source field

hellsgate1001:


hellsgate1001:
Opened in wrong repo. Apologies!

PR #1208, opened 2015-06-08
possible typo fix

ShawnMilo:
It seems that this was meant to be results.


acdha:
Hah, yes – that's been there since 2009! Thanks!

PR #1209, opened 2015-06-08
minor rewording

ShawnMilo:
'several` was misleading because there were only two, and the number wasn't relevant to the point being made.


ShawnMilo:
It's been over four years; looks like it's not going to get merged.

PR #1232, opened 2015-07-23
Rename elasticsearch-py to elasticsearch in docs

dlo:


dlo:
That was fast, thanks!

PR #1234, opened 2015-07-30
need to populate the field map, call super!

yeago:


acdha:
Is there a reason why this doesn't call super() first before running the rest of __init__? I generally prefer to do that in case super() has any side-effects which a subclass needs to override.

Otherwise, looks good to me.

yeago:
@acdha i just saw this. it was never called before this, so I am not sure what side-effects you might be talking about subclassing relying on happening. i also think its likely to be an arbirtary distinction and its perhaps neccessary for something in the large amount of work above that super() to be set for the rest of it to work.

PR #1240, opened 2015-08-21
Use a faster implementation of this cache.

speedplane:
In my tests this new implementation runs much faster and uses far less memory.

Compare the following two:
```
# Takes about a second to run:
sys.getsizeof([None]*pow(10,8))
# Takes much longer and sometimes gets memory errors
sys.getsizeof([None for i in range(pow(10, 8))])
```

This PR simply uses the first implementation instead of the second.

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -175,7 +175,7 @@ class SearchQuerySet(object):
         # an array of 100,000 ``None``s consumed less than .5 Mb, which ought
         # to be an acceptable loss for consistent and more efficient caching.
         if len(self._result_cache) == 0:
-            self._result_cache = [None for i in range(self.query.get_count())]
+            self._result_cache = [None] * self.query.get_count()
 
         if start is None:
             start = 0


phill-tornroth:
This really piqued my interest, so ran my own quick checks, which corroborate the improvement (included the original with xrange vs range because I was curious how big a difference that alone would make).

```
>>> timeit.timeit('[None for i in range(10000)]', number=1000)
0.3423290252685547
>>> timeit.timeit('[None for i in xrange(10000)]', number=1000)
0.3162569999694824
>>> timeit.timeit('[None] * 10000', number=1000)
0.04814600944519043
>>> timeit.timeit('[None for i in range(100000000)]', number=1)
5.295490980148315
>>> timeit.timeit('[None for i in xrange(100000000)]', number=1)
3.534160852432251
>>> timeit.timeit('[None] * 100000000', number=1)
0.8227479457855225
```

So anyways, I buy the change for sure and I can imagine why it's faster (though presumably could be optimized by the runtime someday), but where I'm REALLY stumped is on the memory difference. Can anyone explain why two lists, each containing 100,000,000 None entries are difference sizes!?

```
>>> sys.getsizeof([None]*pow(10,8))
800000072
>>> sys.getsizeof([None for i in range(pow(10, 8))])
859724480
```

Oh, a possibility just occurred to me. Probably the internals of the list are re-sizing a fixed length buffer to make room for new list items. The fast case never has to do this because it's allocated all at once and the slow case has to do this periodically because the list doesn't know when it'll be done getting new kids.. so the extra memory is presumably internal space for adding stuff.

Yeah, that seems to hold true. I don't see the size tick up with new items here:
```
>>> result = [None for i in range(pow(10, 8))]
>>> sys.getsizeof(result)
859724480
>>> for i in xrange(10): result.append(None)
>>> sys.getsizeof(result)
859724480
```

And adding even a single item to the fast result should increase memory if my theory holds:
```
>>> first = [None] * pow(10,8)
>>> sys.getsizeof(first)
800000072
>>> first.append(None)
>>> sys.getsizeof(first)
900000128
```

And damn! Gets way bigger. Presumably the growth function of that structure is based on a percentage of the current size of the list. Anyways, off to go look at that CPython code now that I'm down this far. Apologies for hijacking the thread.

TL/DR: +1 for the PR


(As an aside, calling the existing implementation horrible seems at least a little unnecessarily inflammatory and rude to me.)

speedplane:
You're right about the remark. Edited.

PR #1256, opened 2015-10-15
Create ability for batch level preprocess

ebressler:
Sometimes you need to run some bulk operation before each "batch" is called when in multiprocess mode. (It also works in single process mode as one giant batch).  Examples of this is fetching a bunch of data from a remote system that is costly to do on a model by model basis.  Instead, with a pre-process, you can fetch that data up front and cache it for the life of the batch and just do a memory lookup rather than making costly multiple round trips to an external source.


acdha:
Could you rebase this against the current master and maybe squash the grammar correction patch? The MySQL commit was already merged in #1201 and we'd like to keep the commit history as clean as possible.

ebressler:
Hmm, I thought I did the rebase. I will fix and get a cleaner PR to you.

acdha:
@ebressler: thanks!

ebressler:
@acdha I made a cleaner pr with only the changes I needed.  See if this makes sense now.

ebressler:
@acdha Things good now?

ebressler:
@acdha just pinging things to see if you need anything else from me to merge this.

acdha:
Have you had time to see why the tests are all failing?

ebressler:
@acdha sorry, apparently I was used to them failing from my last submit and didn't check that I typo'd somehow. I must not have pushed a fix up.  I have now. Looks like it is passing for the ones I expect it to now.

ebressler:
@acdha Anything else you need?  Looks all clear now.

haystack/indexes.py
```diff
@@ -353,7 +353,12 @@ def load_all_queryset(self):
         By default, returns ``all()`` on the model's default manager.
         """
         return self.get_model()._default_manager.all()
-
+    
+    def pre_process_data(self, queryset):
```
acdha: I think "preprocess" is the more common spelling

acdha: Also, should we perhaps make the name self-documenting – `preprocess_queryset` or `preprocess_queryset_slice` to convey that it's not the entire queryset?

haystack/management/commands/update_index.py
```diff
@@ -85,6 +85,8 @@ def do_update(backend, index, qs, start, end, total, verbosity=1, commit=True):
     # in memory. Useful when reindexing large amounts of data.
     small_cache_qs = qs.all()
     current_qs = small_cache_qs[start:end]
+    
+    index.pre_process_data(current_qs)
```
acdha: Looking at this API more, I'm inclined to say this should be documented as `current_qs = index.preprocess_queryset(current_qs)` so you could potentially call things like `select_related`, `prefetch_related`, etc. or even replace the queryset entirely if there's a good reason.

haystack/indexes.py
ebressler: I am for changing the names.  Let me look at doing that.

haystack/management/commands/update_index.py
ebressler: Interesting idea, could definitely do that but the purpose was not to alter the queryset for other people to use but rather do some things based on that query set.  I can see the additional use case, but that was not my focus and I hesitate to what to do that.

PR #1258, opened 2015-10-15
Fix #1224 issues

### LINKED ISSUE #1224: Indexing doesn't work with jinja2
Related to #1163  - with jinja2 set as the primary template backend, an exception is raised on `./manage.py update_index`, because the jinja2 render method only accepts a dict, not a `django.template.Context` instance. See https://github.com/django-haystack/django-haystack/blob/be67c37e3d23eed96d675e0c56e85eb6de0d484b/haystack/fields.py#L137

I'm not 100% sure how to solve this in the general case, but simply changing that line to 

    return t.render({'object': obj})

works for both the DTL and jinja2. The downside is no context processors, but they're not likely to be useful in an index template anyway.

coagulant:
TBD, wip

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import re
 
-from django.template import Context, loader
+from django.template import loader
 from django.utils import datetime_safe, six
 
 from haystack.exceptions import SearchFieldError
@@ -134,7 +134,7 @@ class SearchField(object):
             template_names = ['search/indexes/%s/%s_%s.txt' % (app_label, model_name, self.instance_name)]
 
         t = loader.select_template(template_names)
-        return t.render(Context({'object': obj}))
+        return t.render({'object': obj})
 
     def convert(self, value):
         """
--- a/test_haystack/test_fields.py
+++ b/test_haystack/test_fields.py
@@ -4,12 +4,21 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import datetime
 from decimal import Decimal
+from unittest import skipIf
 
 from django.template import TemplateDoesNotExist
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from test_haystack.core.models import MockModel, MockTag
 
 from haystack.fields import *
+
+
+try:
+    import jinja2
+    from django.template.backends.jinja2 import Jinja2
+except ImportError:
+    jinja2 = None
+    Jinja2 = None
 
 
 class CharFieldTestCase(TestCase):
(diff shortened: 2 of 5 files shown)


haystack/fields.py
```diff
@@ -134,7 +135,10 @@ def prepare_template(self, obj):
             template_names = ['search/indexes/%s/%s_%s.txt' % (app_label, model_name, self.instance_name)]
 
         t = loader.select_template(template_names)
-        return t.render(Context({'object': obj}))
+        if django.VERSION >= (1, 8, 0):
```
acdha: Good, this was an issue for #1224

haystack/fields.py
```diff
@@ -134,7 +135,10 @@ def prepare_template(self, obj):
...
+            return t.render({'object': obj})
```
nazarewk: Should probably use `Context({'object': obj}).flatten()` instead (right now it's missing default constants like None, True, False).

for example:
```python
ctx = Context({'object': obj})
if django.VERSION >= (1, 8, 0):
    ctx = ctx.flatten()
t.render(ctx)
```

coagulant: Ok, I'll resubmit this PR with an update, LGTM

galuszkak:
Looks really great! @coagulant did you maybe update this PR with suggestion from @nazarewk (and resolve conflicts?)? IMHO It looks really good!

kkzxak47:
Just ran into this problem today, this PR saved my day. 
But, where are you @coagulant , it's now 2016. =. =

haystack/fields.py
acdha: This check can now be removed since we no longer support Django 1.7 or earlier now that the Django core team has dropped support

haystack/fields.py
acdha: Repeat previous simplification for the new Django 1.8+ support base

acdha:
If someone wants to rebase this is against the current master branch it should be even easier to merge now that we don't need to worry about old versions of Django

nazarewk:
Needs some Django 1.9 fixes for running tests (tons of `ImportError: No module named 'django.db.models.loading'`

acdha:
@nazarewk Django 1.9 tests can be ignored for the moment – I'm merging @claudep's PRs for that now (#1327-1332) and it doesn't look like any of this code actually touches that.

acdha:
@claudep Thanks for this run of pull requests – the tests for the full batch are passing: https://travis-ci.org/acdha/django-haystack/builds/118498371

acdha:
@nazarewk Rebase against 10bcc7668ccaab499f8db516fd59de182aadbcb4 and you should have tests passing on Django 1.9

coagulant:
I've rebased PR against master. Core of the fix was already there in 69bb1db2, so I've left the tests only

test_haystack/test_fields.py
```diff
@@ -4,14 +4,24 @@
...
 
 from haystack.fields import *
 
 
+try:
+    import jinja2
+    from django.template.backends.jinja2 import Jinja2
+except ImportError:
+    jinja2 = None
+    Jinja2 = None
```
claudep: Isn't the first `jinja2` enough to test for Jinja2 availability?

Issue #1190, opened 2015-05-12
Schedule deprecation for Django 1.6 support


acdha:
A fair number of people in the community still use Django 1.6 but it's now two releases behind and officially unsupported by the Django project.

Koed00:
I'll sign up for this one after the 2.4 releasee

simonsmiley:
Since django 1.6 is not used by travis, I guess support was dropped

acdha:
Closing this since we're now following the upstream Django support lifecycle

nitely:
@acdha you forgot to close this one.

acdha:
@nitely thanks!

PR #1260, opened 2015-10-20
Moved signal processor loading to app_config.ready

### LINKED ISSUE #1212: Allow HAYSTACK_SIGNAL_PROCESSOR Class to Import Models
Currently, if you use the `HAYSTACK_SIGNAL_PROCESSOR` function to set to a class which imports a model, then you can get ciruclar imports. This was also reported here:
http://stackoverflow.com/questions/17049426/importerror-when-using-haystack-2-0-0-with-django-1-5-and-gunicorn-wsgi

I am not sure what is causing this, but it would be nice if our signal processors could import our models.

claudep:
Fixes #1212.

### DIFF
--- a/haystack/__init__.py
+++ b/haystack/__init__.py
@@ -2,8 +2,6 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import logging
-
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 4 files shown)


claudep:
As expected, this fails on Django 1.6. Is 1.6 supposed to be supported after the 2.4 release or not?

acdha:
@claudep I'm leaning no unless someone has a really compelling reason (i.e. they're willing to step up and support it). The core Django team dropped support for 1.6 in April and I see no reason to support it longer than they do: https://www.djangoproject.com/download/

claudep:
Dropping 1.6 is tracked in #1190. I guess most of the work is on the documentation side.

acdha:
Likely, yes, although it might be a good time to review the various conditional imports & wrapper functions. If memory serves, a bunch of those are moot once we only support Django 1.7+.

claudep:
Rebased on current master.

PR #1261, opened 2015-10-24
Rename 'page_obj' with 'page' in the tutorial

Beatris:
Rename 'page_obj' with 'page' in the tutorial, section Search Template as there is no 'page_obj' in the "django-haystack/haystack/views.py" and this results giving 'No results found' in the search.


acdha:
Thanks for the pull-request – I just merged your corrections.

PR #1265, opened 2015-11-04
The facets size can be now set via a setting.

jmiguelv:
The setting HAYSTACK_ELASTICSEARCH_FACETS_SIZE can be used to set the number of facets to be returned by elasticsearch.

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -351,7 +351,7 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
                 facet_options = {
                     'terms': {
                         'field': facet_fieldname,
-                        'size': 100,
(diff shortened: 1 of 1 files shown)


SalahAdDin:
:+1:

sampeka:
If this is not supported already, it would probably be better to integrate this on the SearchQuerySet api, similiar to how Solr has support at the moment. The SearchQuerySet api supports setting the facet size on Solr using the `SearchQuerySet.facet` method, e.g. `sqs.facet(my_field, limit=100)`. Does the same not work for elasticsearch, e.g. `sqs.facet(my_field, size=100)`? 

See the docs here: http://django-haystack.readthedocs.org/en/latest/faceting.html#configuring-facet-behaviour

I'm going to close this PR in it's current format, but please feel free to submit a new one that improves support for elasticsearch.

PR #1268, opened 2015-11-11
Fix import error for elasticsearch >= 1.6.0

ahmontero:
changed import to fix compatibility issue with elasticsearch >= 1.7.0, because 'bulk_index' was deprecated from elasticsearch > 1.6.0

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -20,8 +20,11 @@ from haystack.utils import get_identifier, get_model_ct
 
 try:
     import elasticsearch
-    from elasticsearch.helpers import bulk_index
     from elasticsearch.exceptions import NotFoundError
+    try:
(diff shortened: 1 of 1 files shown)


acdha:
Fixed in a2c1a06852dfa208ef90d3d3f2b000be2e5cca82 using the patch from #1254 which is in the master branch. I would recommend upgrading from v2.3 as soon as possible.

iMerica:
My `requirements.txt` specifies `django-haystack==2.4.0` and I'm getting the same error. Are there any other workarounds?

acdha:
@iMerica The latest version is 2.4.1, which includes that patch:

https://pypi.python.org/pypi/django-haystack/2.4.1
https://github.com/django-haystack/django-haystack/commit/340ce0f6a5abbb7c887ffa5e7fd2146d061a0a77

PR #1269, opened 2015-11-11
Fix queryset slicing and code duplication

### LINKED ISSUE #370: RelatedSearchQueryset loads lots of pages from backend when paginated
We found that hitting 

```
/mymodel/?page=6
```

was causing haystack to hit solr for each page until it got to the right one. That is, it would hit all these urls:

```
/solr/select?...&page=1
/solr/select?...&page=2
/solr/select?...&page=3
/solr/select?...&page=4
/solr/select?...&page=5
/solr/select?...&page=6
```

Needless to say that's a bit crap. I found the bug - looks like the **getitem** and _fill_cache methods on RelatedSearchQueryset don't quite behave themselves in a rather bizarre fashion. I merged some of the diffs from SearchQueryset and it fixed the problem.

Here's the fix: https://github.com/koordinates/django-haystack/commit/8fe85084a707edf6fee9f83526e41e79c0084f56

Ideally I would extract some of the behaviour from _fill_cache into a separate method so RelatedSearchQueryset doesn't have to override the whole huge method, but I don't know what the best way to do that would be so I left it for now.

craigds:
Now to get to `?page=100`, you don't have to iterate all 100 pages from the backend. Also removes a lot of duplicate code from RelatedSearchQuerySet.

Fixes #370.

replaces #960, which was my previous attempt at solving this but it's ancient now and I never bothered to get all the tests passing (and there were a couple of bugs.)

I've removed the `_fill_cache` test because I couldn't see a way to get it working, and no code actually calls `_fill_cache` in the manner the test does, so it seemed a bit irrelevant. Plus it's called by methods that are covered by other tests anyway.

One thing - if you're using a `load_all_queryset` and some of the results from the search backend are not in the `load_all_queryset`, some of the results at page 100 may end up being the same as those at page 99. That's because the page 99 will skip those results and fill the gaps by loading some more instead (and they'll be the same ones that page 100 will load). If you have a _lot_ of this happening, you'll probably notice a lot of dupe results.

However, IMHO it beats the previous behaviour of never actually loading page 100 (because RelatedSearchQuerySet spent ages iterating 100 pages of results first, eventually causing a page timeout.)

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -218,11 +218,19 @@ class SearchQuerySet(object):
     def _fill_cache(self, start, end, **kwargs):
         # Tell the query where to start from and how many we'd like.
         self.query._reset()
-        self.query.set_limits(start, end)
-        results = self.query.get_results(**kwargs)
 
         if start is None:
             start = 0
+
+        query_start = start
+        query_start += self._ignored_result_count
+        query_end = end
+        if query_end is not None:
+            query_end += self._ignored_result_count
+
+        self.query.set_limits(query_start, query_end)
+        results = self.query.get_results(**kwargs)
+
 
         if results is None or len(results) == 0:
             # trim missing stuff from the result cache
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -178,15 +178,7 @@ class SearchQuerySet(object):
 
             # Load the objects for each model in turn.
             for model in models_pks:
-                try:
-                    ui = connections[self.query._using].get_unified_index()
-                    index = ui.get_index(model)
-                    objects = index.read_queryset(using=self.query._using)
-                    loaded_objects[model] = objects.in_bulk(models_pks[model])
-                except NotHandled:
-                    self.log.warning("Model '%s' not handled by the routers", model)
-                    # Revert to old behaviour
-                    loaded_objects[model] = model._default_manager.in_bulk(models_pks[model])
+                loaded_objects[model] = self._load_model_objects(model, models_pks[model])
 
         for result in results:
             if self._load_all:
@@ -203,6 +195,9 @@ class SearchQuerySet(object):
                     # The object was either deleted since we indexed or should
                     # be ignored; fail silently.
                     self._ignored_result_count += 1
+
+                    # avoid an unfilled None at the end of the result cache
+                    self._result_cache.pop()
                     continue
 
             to_cache.append(result)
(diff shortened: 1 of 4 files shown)


sampeka:
Not sure if removing `test_fill_cache` is a good idea, especially since it's not replaced by anything. Furthermore, from my perspective the test case looked reasonable - why was it failing? Considering that this is quite a significant change to a core class, I'm not sure this is mergeable in its current state.

> One thing - if you're using a load_all_queryset and some of the results from the search backend are not in the load_all_queryset, some of the results at page 100 may end up being the same as those at page 99. That's because the page 99 will skip those results and fill the gaps by loading some more instead (and they'll be the same ones that page 100 will load). If you have a lot of this happening, you'll probably notice a lot of dupe results.

Could you explain a little bit more by what you mean by this?

craigds:
> Could you explain a little bit more what you mean by this?

Previous behaviour: at ?page=100, django-haystack was going to pysolr and asking for:
```
/solr/?q=foo?page=1
/solr/?q=foo?page=2
/solr/?q=foo?page=3
...
/solr/?q=foo?page=100
```

This was really slow and awful. I think the reasoning was that, if solr returned a result that no longer exists in the database, we could skip it, and the results we'd display would be consistent across all pages.

To clarify, suppose 
`/solr/?q=foo?page=99` has results 990, 991, 992, 993, 994, 995, 996, 997, 998, 999.
`/solr/?q=foo?page=100` has results 1000, 1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, 1009.

If result 995 has actually been deleted from the database, previous behaviour was:

```
?page=99: [990, 991, 992, 993, 994, 996, 997, 998, 999, 1000]
?page=100: [1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, 1009, 1010]
```

New behaviour is this:

```
?page=99: [990, 991, 992, 993, 994, 996, 997, 998, 999, 1000]
?page=100: [1000, 1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, 1009]
```

Since the code no longer hits the backend for every previous page, it can no longer re-align the pages. So result 1000 appears on both pages.

> why was it failing?

Probably fair enough. I can't remember, but I can look into it if you're happy with the above behaviour change.

sampeka:
I'll have a look at this hopefully sometime over the next few days and get back to you asap.

SalahAdDin:
:+1:

Issue #1267, opened 2015-11-09
Elasticsearch engine not working on Django 1.9b1 


tarvitz:
Standard (tutorial) work-around with elasticsearch engine selected does not work on django-1.9b1 though 1.8.6 works well.

Still have some `removed in django-1.9b1` parts such as models.get_model (haystack.forms:111)

acdha:
Django 1.9 is not supported yet. Please open a pull request with any code
changes you've made which are not already in the existing pull requests.

mariocesar:
@tarvitz I made a fix in this PR #1276

tarvitz:
@mariocesar, thanks I'll try to test my workaround with this pr a litte bit later.

PR #1276, opened 2015-11-24
Use compatible get_model util to support new django versions

mariocesar:
For the Django 1.9 version it fails to load the model property in the Result Class, this due`model.get_model` is deprecated in favor of the new `app.get_model`.

As django-haystack already haves a compat method to load model classes I'm just reusing this.


mariocesar:
This fixes #1267

jungornti:
Highly waiting for new release.

PR #1281, opened 2015-12-06
Add python 3.5 to tests

itbabu:
Django 1.8 supports python 3.5
see: https://docs.djangoproject.com/en/1.9/faq/install/#what-python-version-can-i-use-with-django


bigjust:
Thanks!

PR #1292, opened 2016-01-07
import connection_router in SearchQuerySet's method rather in header

rage2000:
import connection_router in SearchQuerySet's method rather in header fix a bug when we use a specific router (if not haystack will use the default router)

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -41,7 +41,7 @@ class SearchQuerySet(object):
         self.log = logging.getLogger('haystack')
 
     def _determine_backend(self):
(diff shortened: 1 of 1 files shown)


sampeka:
Could you describe the bug this PR is solving? What benefit does importing `connection_router` in the body of the method have when its being imported at the top of the file?

sampeka:
Could you remove the redundant `connection_router` import from the top of the file?

sampeka:
Closed in favour of https://github.com/django-haystack/django-haystack/pull/1296

Issue #1001, opened 2014-05-12
solr_tests test_multiprocessing sometimes fail


jorgecarleitao:
This test is making some pull requests to fail the travis, which is making it more difficult to properly evaluate them.

Examples:
* #951, [this build](https://travis-ci.org/toastdriven/django-haystack/jobs/21842097#L1359)
* #978, [this build](https://travis-ci.org/toastdriven/django-haystack/jobs/21307417#L643)

For now, this is just documenting the occurrence.

acdha:
How tedious: 1ab44b56c533c4cc262d543b9794aa7ddd2d63c6 shows every sign of fixing this issue (tested with a few thousand repeats of that test) but it appears to hang on Python 2.6's multiprocessing:

https://travis-ci.org/acdha/django-haystack/builds/48965711

Time for more testing and maybe just disabling parallel workers on 2.6…

PR #1296, opened 2016-01-15
Load routers lazily (fixes #1034)

### LINKED ISSUE #1034: connection_router is not actually a singleton when custom routers are used
In haystack/__init__.py a pattern is employed where connection_router is assigned to a ConnectionRouter instance with no custom routers, and then if settings says so we'll create a ConnectionRouter with the custom classes.

Problem is, ConnectionRouter constructs those custom Router classes.. and there's a pretty good chance they live in modules that (directly or indirectly) import some haystack stuff (lots of my haystack customization was in the same module of mine).

When that happens it's likely that things like haystack/indexes are gonna do a `from haystack import connections, connection_router` and permanently end up with the empty connection_router instance, which the stack is currently attempting to replace.

In my eyes, the only really solid solution is to make it impossible for there to be two instances assigned to haystack.connection_router at different times. The most reasonable way I came up with for doing that is to lazily load the router classes at first-use.

Here's the commit on my fork, which is branched off of v2.0.0 at the moment. I'm happy to open up a pull request also but the routing functionality seems potentially experimental and without tests (unless I've overlooked them). So figured I'd open the issue first instead and see what folks think.

https://github.com/phill-tornroth/django-haystack/commit/12536f4ee50c13786f70343dde72391a017212f7

skirsdeda:
This moves connection_router import in ```SearchQuerySet``` to ```_determine_backend``` (makes it local import) which already fixes the problem. But to make it fail-proof, router loading is now lazy. **HAYSTACK_ROUTERS** setting is still checked for validity in **\__init\__.py**

### DIFF
--- a/test_haystack/test_loading.py
+++ b/test_haystack/test_loading.py
@@ -97,6 +97,11 @@ class ConnectionRouterTestCase(TestCase):
         cr = loading.ConnectionRouter()
         self.assertEqual([str(route.__class__) for route in cr.routers], ["<class 'haystack.routers.DefaultRouter'>"])
 
+    @override_settings(HAYSTACK_ROUTERS=[])
+    def test_router_override1(self):
+        cr = loading.ConnectionRouter()
+        self.assertEqual([str(route.__class__) for route in cr.routers], ["<class 'haystack.routers.DefaultRouter'>"])
+
     @override_settings(HAYSTACK_ROUTERS=['test_haystack.mocks.MockMasterSlaveRouter', 'haystack.routers.DefaultRouter'])
     def test_router_override2(self):
         cr = loading.ConnectionRouter()
--- a/test_haystack/test_loading.py
+++ b/test_haystack/test_loading.py
@@ -2,6 +2,7 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.test import TestCase, override_settings
 from test_haystack.core.models import AnotherMockModel, MockModel
@@ -88,7 +89,9 @@ class ConnectionHandlerTestCase(TestCase):
 
 
 class ConnectionRouterTestCase(TestCase):
+    @override_settings()
     def test_init(self):
+        del settings.HAYSTACK_ROUTERS
         cr = loading.ConnectionRouter()
         self.assertEqual([str(route.__class__) for route in cr.routers], ["<class 'haystack.routers.DefaultRouter'>"])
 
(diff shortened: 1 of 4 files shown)


skirsdeda:
The only downside of this is that if you specify invalid path to a router class in ```HAYSTACK_ROUTERS```, this will only show up on first ```SearchQuerySet``` instantiation. But I guess that's not a big problem at all.

skirsdeda:
@sampeka not much difference. ```_routers``` is only accessed in ```def routers```.

Also fixed tests and added handling of ```HAYSTACK_ROUTERS=[]```.

skirsdeda:
@acdha Could you take a look at build log? To me it seems that python 3.5 and pypy build fails are unrelated to this PR.

acdha:
@skirsdeda Indeed – there's a race condition in the multiprocessing test which has been tedious to resolve because it generally only reproduces on relatively slower servers like the Travis build farm: https://github.com/django-haystack/django-haystack/issues/1001. I just restarted that test job.

sampeka:
@skirsdeda This looks good :+1: , I'm definitely happy for this to be merged in if you could move the instance variable to be set in `__init__`. I know it's not a big deal since it's a tiny class, but it's just better practice to define instance variables in a single place and it also means you don't have to use `getattr` later on. That is me being quite anal though ;)

skirsdeda:
@acdha Do you think this is good to go in?

test_haystack/test_loading.py
```diff
@@ -88,47 +89,48 @@ def test_get_unified_index(self):
 
 
 class ConnectionRouterTestCase(TestCase):
+    @override_settings()
```
acdha: Is there a reason to have `override_settings` with an empty list?

acdha: Never mind, just noticed https://code.djangoproject.com/ticket/20032.

acdha:
@skirsdeda Looks reasonable to me – I haven't run it but the change seems reasonable to me. The only thing I noticed is future-proofing paranoia – since we document the router list as accepting either lists or tuples, perhaps we should have a test which uses a tuple just to avoid a future change breaking that contract?

skirsdeda:
@acdha I see. How about adding something like ```RemovedInHaystack27``` warning (just like in Django). Then I could readd routers_list arg and emit this warning when it's used. This seems like a better option to me because junk like this is accumulating all the time and when it's never removed further development becomes really tedious.

We could also have a deprecation timeline (like https://docs.djangoproject.com/en/1.9/internals/deprecation/).

acdha:
If nothing else, we should add a DeprecationWarning – retroactively if you
notice an area where we could remove a now/soon-to-be stale chunk of code.

PR #1297, opened 2016-01-18
Highlight elasticsearch 2.X is not supported yet

martinsvoboda:
I spend several hours with problems with elasticsearch 2

### DIFF
--- a/docs/backend_support.rst
+++ b/docs/backend_support.rst
@@ -50,7 +50,7 @@ Elasticsearch
 * Stored (non-indexed) fields
 * Highlighting
 * Spatial search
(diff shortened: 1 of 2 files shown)


sampeka:
Good catch :+1:  there's also a reference to the compatible versions of elasticsearch in `docs/installing_search_engines.rst`. If you could fix that as well I'll merge it in

martinsvoboda:
Done, it is ready for merge.

sampeka:
Thanks, final thing - can you squash the commits? Better to keep the commit history as clean as possible.

http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed

martinsvoboda:
Sound reasonable. Done.

PR #1302, opened 2016-01-21
Retry backend updates

giladbeeri:
added: retries when backend update fails.

Sometimes, on a long update_index command, the backend raises an exception which aborts the whole process. A lot of times it's a ConnectionError which is temporary. Added retry attempts to reduce failure chances


haystack/management/commands/update_index.py
```diff
@@ -86,14 +87,35 @@ def do_update(backend, index, qs, start, end, total, verbosity=1, commit=True):
...
+        try:
+            # FIXME: Get the right backend.
+            backend.update(index, current_qs, commit=commit)
+            if verbosity >= 2 and retries:
+                print('Succeeded indexing {} - {}, tried {}/{} times'.format(start + 1, end, retries + 1, max_retries))
+            break
+        except:
+            retries += 1
+            if verbosity >= 2:
+                exc = traceback.format_exc()
```
acdha: I would prefer to use `logging.exception` (or `logging.warning` with `exc_info=True`) for this since it avoids the need for custom formatting logic, sending errors to `stderr` instead of `stdout`, etc. 

In this case I'd also like to avoid the blanket except in favor of catching a specific category of exception (e.g. `EnvironmentError`, `IOError`, etc.) so things like `KeyboardInterrupt` and `SystemExit` still work.

acdha:
I think this would be more useful with some kind of back-off mechanism so if, for example, the error happened because a daemon was just starting up it wouldn't rapidly hit the retry count and fail when simply delaying a second or two would have worked. 

Maybe something as simple as tossing a `time.sleep(2 ** retries)` after displaying the retry message?

giladbeeri:
@acdha I referenced all of your comments + added the option to configure `max_retries` from the command line.
Catching Exception as this is the base class for Elasticsearch's exceptions + SystemExit and KeyboardInterrupt do not inherit from it.

giladbeeri:
@acdha  Any news about that?

acdha:
Can you rebase against master? I should be able to take care of merging this week.

Chris

giladbeeri:
done

PR #1148, opened 2015-02-27
Adaptations for Solr 5.0.0

elishowk:
Solr 5.0.0 is easy to deploy with cloud Cores with automatic managed-schema and no more schem.xml.
Solr 5.0.0 can edit a core's configuration via with the Schema REST API. 
This is what this pull request does. Along with an updated documentation.

Refers to issue #1147 and try to resolve it...

PR #1304, opened 2016-01-29
Allow SearchField objects to set Solr field type

ahoho:
Currently there's no easy way for a given `SearchField` object to explicitly set its Field Type in the Solr Schema. Obviously, this can be changed from within the Solr Admin page, but I feel that an integrated solution is preferable.

This is particularly useful for indexed fields that need not be tokenized and should support sorting (e.g., ID numbers, emails, etc.).

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -26,7 +26,7 @@ class SearchField(object):
     def __init__(self, model_attr=None, use_template=False, template_name=None,
                  document=False, indexed=True, stored=True, faceted=False,
                  default=NOT_PROVIDED, null=False, index_fieldname=None,
-                 facet_class=None, boost=1.0, weight=None, solr_explicit_field_type=''):
+                 facet_class=None, boost=1.0, weight=None, solr_explicit_field_type=None):
         # Track what the index thinks this field is called.
         self.instance_name = None
         self.model_attr = model_attr
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -26,7 +26,7 @@ class SearchField(object):
     def __init__(self, model_attr=None, use_template=False, template_name=None,
                  document=False, indexed=True, stored=True, faceted=False,
                  default=NOT_PROVIDED, null=False, index_fieldname=None,
-                 facet_class=None, boost=1.0, weight=None, solr_explicit_field_type=None):
+                 facet_class=None, boost=1.0, weight=None, solr_field_params=None):
         # Track what the index thinks this field is called.
         self.instance_name = None
         self.model_attr = model_attr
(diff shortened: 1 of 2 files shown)


acdha:
We should probably add some tests for this. 

I'm also wonder whether the attribute should either be clearly Solr-specific, be implemented for at least ElasticSearch, or perhaps just issue a warning / NotImplementedError when encountered by the non-Solr backends.

ahoho:
Those are a good points. I'm not sure how to handle tests, but I'll make the changes to either specify Solr-specific compatibility or add support for Elasticsearch.

On Jan 29, 2016, at 3:50 PM, Chris Adams <notifications@github.com<mailto:notifications@github.com>> wrote:


We should probably add some tests for this.

I'm also wonder whether the attribute should either be clearly Solr-specific, be implemented for at least ElasticSearch, or perhaps just issue a warning / NotImplementedError when encountered by the non-Solr backends.

—
Reply to this email directly or view it on GitHub<https://github.com/django-haystack/django-haystack/pull/1304#issuecomment-176964075>.
******************
 
IMPORTANT NOTICE FROM THE BRATTLE GROUP: This message, and any attachments, are
intended only for the use of the individual or entity to which it is addressed and
may contain information that is privileged, confidential, and exempt from disclosure
under applicable law. Any unauthorized dissemination, distribution or copying of this
communication is strictly prohibited. If you have received this communication in error
please notify me immediately by return email and immediately delete the original and all
copies of the message and any attachments to it. Note also that nothing in this message is
intended to constitute an electronic signature or otherwise to satisfy the requirements for
a contract unless an express statement to the contrary is included in the message.

Please ensure you have adequate virus protection before you open or use attachments.
The Brattle Group does not accept any liability for viruses.
 
******************

ahoho:
@acdha, this patch now raises a `NotImplementedError` when using a non-Solr backend. I'm still now sure how best to test (outside of confirming that it works on my own project).

haystack/fields.py
```diff
@@ -26,7 +26,7 @@ class SearchField(object):
     def __init__(self, model_attr=None, use_template=False, template_name=None,
                  document=False, indexed=True, stored=True, faceted=False,
                  default=NOT_PROVIDED, null=False, index_fieldname=None,
-                 facet_class=None, boost=1.0, weight=None):
+                 facet_class=None, boost=1.0, weight=None, solr_explicit_field_type=''):
```
acdha: Can we use None for this to match the other optional parameters?

ahoho:
@acdha, I've made your suggested change.

I know there is limited manpower and far more important pull requests and issues to review, so I'll gladly close this to lighten the load if you think it's an unnecessary fix. My feeling was/is that, despite talk in PR #1148 that suggests the use of the Rest API, it seems like good practice to have `search_indexes.py` match whatever is going on in the schema to the extent possible.

ahoho:
Changed to be a more general fix -- now the user can make direct alterations to any field in the Solr schema via a `solr_field_params` argument. Now additional changes to fields can be passed as a dictionary, e.g.,

 `my_field = indexes.SearchField(solr_field_params = {'type': 'string', 'sortMissingFirst': 'true'})`

PR #1313, opened 2016-02-17
Remove boost which is now unsupported

chrisbrooke:
In Elasticsearch >2.0, boost is no longer a support meta-data:

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#migration-meta-fields

See: http://stackoverflow.com/questions/35443179/django-haystack-locationfield-created-as-string-instead-of-geo-point-in-elastics

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -135,10 +135,6 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
         current_mapping = {
             'modelresult': {
                 'properties': field_mapping,
-                '_boost': {
-                    'name': 'boost',
-                    'null_value': 1.0
-                }
             }
         }
 


acdha:
@chrisbrooke Are there any downsides to doing this for ElasticSearch <2.0, which is still pretty common (e.g. AWS's ElasticSearch service runs 1.5)

chrisbrooke:
@acdha according to the docs _boost has been deprecated in Elasticsearch as of v1.0.0.RC1 (https://www.elastic.co/guide/en/elasticsearch/reference/1.4/mapping-boost-field.html#function-score-instead-of-boost). Boost can still be supplied on a field level by adding the boost value to the seach_index

acdha:
Perfect – thanks for checking!

Issue #188, opened 2010-02-24
Faceted data should occur in ``post_prepare``


toastdriven:
Move the data population to a new `post_prepare` hook so subclasses don't have to do stupid tricks to make faceting work on data they're custom preparing.

PR #1312, opened 2016-02-17
haystack with rpyc (RPC)

acdha:
rpyc 3.3.0
haystack 2.0
python 2.7

Conflict in the method `__getattr__`

[rpyc issue #188](https://github.com/tomerfiliba/rpyc/issues/188)


acdha:
This is still a work in progress but leads to a number of areas for cleaning up old code

Issue #1069, opened 2014-09-14
Update test mocks to deal with Django 1.7's registry


acdha:
Currently two tests fail because they rely on mock models which aren't registered:

```
======================================================================
ERROR: test_missing_object (test_haystack.test_models.SearchResultTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/toastdriven/django-haystack/test_haystack/test_models.py", line 137, in test_missing_object
    self.assertEqual(awol2.model, None)
  File "/home/travis/build/toastdriven/django-haystack/haystack/models.py", line 98, in _get_model
    self._model = models.get_model(self.app_label, self.model_name)
  File "/home/travis/build/toastdriven/django-haystack/.tox/py27-django1.7/lib/python2.7/site-packages/django/db/models/__init__.py", line 54, in alias
    return getattr(loading, function_name)(*args, **kwargs)
  File "/home/travis/build/toastdriven/django-haystack/.tox/py27-django1.7/lib/python2.7/site-packages/django/apps/registry.py", line 202, in get_model
    return self.get_app_config(app_label).get_model(model_name.lower())
  File "/home/travis/build/toastdriven/django-haystack/.tox/py27-django1.7/lib/python2.7/site-packages/django/apps/config.py", line 161, in get_model
    "App '%s' doesn't have a '%s' model." % (self.label, model_name))
LookupError: App 'core' doesn't have a 'yetanothermockmodel' model.
-------------------- >> begin captured logging << --------------------
haystack: ERROR: Object could not be found in database for SearchResult '<SearchResult: core.mockmodel (pk='1000000')>'.
--------------------- >> end captured logging << ---------------------
======================================================================
FAIL: test_more_like_this_for_model (test_haystack.solr_tests.test_templatetags.MoreLikeThisTagTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/toastdriven/django-haystack/mock-1.0.1-py2.7.egg/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/toastdriven/django-haystack/test_haystack/solr_tests/test_templatetags.py", line 55, in test_more_like_this_for_model
    any_order=True)
  File "/home/travis/build/toastdriven/django-haystack/mock-1.0.1-py2.7.egg/mock.py", line 877, in assert_has_calls
    '%r not all found in call list' % (tuple(not_found),)
[...]

PR #1322, opened 2016-03-09
Maintenance updates and fixes for v2.3 branch.

aramboi:
Backported the following fixes from 2.4.x:
- Added source encoding declaration header. (#be67c37)
- Adding fix for previously test for django 1.7. (#99fb662)

Fixed Solr version and download path for travis build tests.
Fixed requirement for elasticsearch-py<2.0.0 in setup.py.
Fixed tests failing with latest 1.x version of elasticsearch-py.
Updated README to show the proper build image and link.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ MANIFEST
 env
 env3
 *.egg
+.eggs
 .coverage
 
 test_haystack/solr_tests/server/solr-4.6.0.tgz
--- a/README.rst
+++ b/README.rst
@@ -43,8 +43,8 @@ Documentation
 Build Status
 ============
 
-.. image:: https://travis-ci.org/toastdriven/django-haystack.svg?branch=master
-   :target: https://travis-ci.org/toastdriven/django-haystack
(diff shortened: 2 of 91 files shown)


aramboi:
This job https://travis-ci.org/django-haystack/django-haystack/jobs/114827437 failed probably because of a network issue while downloading Solr. This happened to me before with travis builds, If you rerun the job it should be successful.

The Django 1.7 build fails in one test due to a know issue: https://github.com/django-haystack/django-haystack/issues/1069 - not in the scope of this PR.

The PyPy build fails are not in the scope of this PR.

This PR should fix 8 out of the 13 Travis CI builds compared to previous build of this maintenance branch.

acdha:
@aramboi Many thanks for digging into that failure - this looks good to me since we already know that 2.3.x isn't committed to supporting new versions of Django.

aramboi:
@acdha I would look into the PyPy failures also, but I don't have much experience with PyPy.
This PR fixes the encoding issue I had while deploying this version on Heroku and should provide a more stable release for people still stuck with Django 1.5 for whatever reason.

PR #1326, opened 2016-03-18
Index obj.pk as integer

ketanbhatt:
Allows for ordering of models efficiently

### DIFF
--- a/haystack/indexes.py
+++ b/haystack/indexes.py
@@ -194,7 +194,7 @@ class SearchIndex(with_metaclass(DeclarativeMetaclass, threading.local)):
         self.prepared_data = {
             ID: get_identifier(obj),
             DJANGO_CT: get_model_ct(obj),
-            DJANGO_ID: force_text(obj.pk),
+            DJANGO_ID: obj.pk,
         }
 
         for field_name, field in self.fields.items():
--- a/haystack/indexes.py
+++ b/haystack/indexes.py
@@ -15,11 +15,6 @@ from haystack.fields import *
 from haystack.manager import SearchIndexManager
 from haystack.utils import get_facet_field_name, get_identifier, get_model_ct
 
-try:
-    from django.utils.encoding import force_text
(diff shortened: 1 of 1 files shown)


ahoho:
Models are already ordered by the integer pk during indexing, as seen on [this line](https://github.com/django-haystack/django-haystack/pull/1326/commits/a133987687ba51224c5bdebe025ef591da798104#diff-d85dbf0b89622849b952ea0c394a8a8bR188). What's the impetus behind this change?

ketanbhatt:
I am using haystack for my admin searches. The objects are indexed according to pk, but not returned in that order when I do a search. 
Now if I want to sort the objects returned in the reverse order, I can't without creating a separate field to store the `pk`, but that will mean indexing the same information twice. 

Also, if we are storing the `django_id`, storing it as a string makes it unusable. If we are storing it, why not store it in a format we can use it.

acdha:
@ketanbhatt Django doesn't require the `pk` to be an integer and thus Haystack cannot be any stricter. If you're absolutely certain that you will never use a non-integer value for any of the models which you're using, you can change the schema to specify an integer type. I personally would recommend ordering using a human-meaningful field instead of an autogenerated ID.

ketanbhatt:
@acdha not that good with django, but you can use a non-integer pk? 
When I search on the admin, I want them to be ordered by PK by default. 

Keeping it a string doesn't help me.

ketanbhatt:
By pk I mean the ID.

acdha:
@ketanbhatt Django allows you to set `primary_key` on almost any field – see https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.Field.primary_key – which includes string values. UUIDs are a common choice for some people since that allows multiple nodes to create values independently without ever generating a conflict.

If you want to order them in the admin, you can change your search schema to use an integer instead of a string and that will be fine as long as you never attempt to index a non-integer value.

PR #1327, opened 2016-03-19
Simplified urls.py

claudep:

### DIFF
--- a/haystack/urls.py
+++ b/haystack/urls.py
@@ -4,13 +4,9 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from haystack.views import SearchView
 
-try:
-    from django.conf.urls import patterns, url
-except ImportError:
-    from django.conf.urls.defaults import patterns, url
+from django.conf.urls import url
 
 
-
-urlpatterns = patterns('haystack.views',
+urlpatterns = [
     url(r'^$', SearchView(), name='haystack_search'),
-)
+]
--- a/test_haystack/core/urls.py
+++ b/test_haystack/core/urls.py
@@ -2,27 +2,26 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from django.conf.urls import include, patterns, url
+from django.conf.urls import include, url
 from django.contrib import admin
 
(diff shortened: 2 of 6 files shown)


haystack/urls.py
```diff
@@ -1,16 +1,12 @@
 # encoding: utf-8
 
-from __future__ import absolute_import, division, print_function, unicode_literals
```
acdha: This doesn't seem like a necessary change. I'd like to leave that in so we have a consistent __future__ policy across every file.

claudep: Yes, I've seen that afterwards. I might suggest a cleaning patch later, as some of these imports are no longer necessary with Python 2.7.

test_haystack/core/urls.py
```diff
@@ -1,28 +1,26 @@
...
+from __future__ import unicode_literals
 
-from django.conf.urls import include, patterns, url
+from django.conf.urls import include, url
 from django.contrib import admin
 
+from haystack import views
 from haystack.forms import FacetedSearchForm
 from haystack.query import SearchQuerySet
-from haystack.views import FacetedSearchView, SearchView
```
acdha: I would prefer to revert this change because it allows static linters to verify that referenced variables exist without having to load views.py

test_haystack/core/urls.py
```diff
@@ -1,28 +1,26 @@
...
 
 
-urlpatterns = patterns('',
-    (r'^admin/', include(admin.site.urls)),
-)
+urlpatterns = [
+    url(r'^admin/', include(admin.site.urls)),
 
+    url(r'^$', views.SearchView(load_all=False), name='haystack_search'),
+    url(r'^faceted/$', views.FacetedSearchView(searchqueryset=SearchQuerySet().facet('author'),
```
acdha: This should probably be indented to match the `FacetedSearchView(` so it's visually obvious which parameters are being passed to which function.

acdha:
Other than the very minor things I added comments for, this looks good to me — there's plenty of room for this now that we don't have to support anything newer than Django 1.8.

PR #1334, opened 2016-03-22
enable .highlight(backend_specific_options)

tymofij:
ElasticSearch (and may be other backends) lets user customise tags used for highlighting search results. This is important for highlighting HTML texts: they can contain `em` already. Let's expose this functionality as an option that can be passed to SearchQuerySet.highlight()

Example usage:
```
result = query.filter('needle').highlight({'pre_tags': ['<i>'],'post_tags': ['</i>']})
```


acdha:
@tymofij This looks good to me – it seems like the only other thing we might want to do is look into documenting the option and add a separate issue for Solr support, since there are quite a few configurable options there even though they're all quite different.

PR #1341, opened 2016-03-26
autoprefix 'hl.' for Solr options

tymofij:
All Solr highlighting options start with "hl.", so let's add this prefix automatically.
https://wiki.apache.org/solr/HighlightingParameters

This will make option dicts shorter: `{'maxAnalyzedChars': 42}` instead of `{'hl.maxAnalyzedChars': 42'}`
And let some of the options be used as simple keyword arguments in QuerySet: `.highlight(preserveMulti=False)`

@acdha

### DIFF
--- a/haystack/backends/solr_backend.py
+++ b/haystack/backends/solr_backend.py
@@ -188,7 +188,10 @@ class SolrSearchBackend(BaseSearchBackend):
             kwargs['hl.fragsize'] = '200'
 
             if isinstance(highlight, dict):
-                kwargs.update(highlight)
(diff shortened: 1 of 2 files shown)


acdha:
My thought was that we should allow the same values from the Solr docs or a test URL. Maybe we should check for `hl.` and only add if it's not present so it'll work either way?

tymofij:
Makes sense. Updated the PR.

SteveByerly:
Do you want to update those docs that I just updated yesterday :)

PR #1358, opened 2016-05-09
Make elasticsearch update respect silently_fail

ErwinJunge:
The update method of the elasticsearch backend doesn't fully respect the silently_fail boolean. This fixes that.

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -190,10 +190,20 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
                                extra={"data": {"index": index,
                                                "object": get_identifier(obj)}})
 
-        bulk(self.conn, prepped_docs, index=self.index_name, doc_type='modelresult')
+        try:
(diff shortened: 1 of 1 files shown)


haystack/backends/elasticsearch_backend.py
```diff
@@ -190,10 +190,20 @@ def update(self, index, iterable, commit=True):
                                extra={"data": {"index": index,
                                                "object": get_identifier(obj)}})
 
-        bulk(self.conn, prepped_docs, index=self.index_name, doc_type='modelresult')
+        try:
+            bulk(self.conn, prepped_docs, index=self.index_name, doc_type='modelresult')
+        except elasticsearch.TransportError as e:
+            if not self.silently_fail:
+                raise
+            self.log.error(u'Error bulk updating Elasticsearch')
```
acdha: Can we update both of these calls to add the index name, set `exc_info=True`, and see if there's any helpful context to pass to logging? Basically using the call on line 189-192 above as an example?

ErwinJunge:
@acdha Done.

PR #1362, opened 2016-05-15
Support ManyToManyFields in model_attr lookups

Terr:
Hi,

This pull request adds support for ManyToManyFields in SearchFields' `model_attr` attribute lookups.

Before the change, a `model_attr` with value `related_models__name` (where `related_models` is a `ManyToManyField`) would result in the following exception:

    haystack.exceptions.SearchFieldError: The model '<ManyToManyLeftSideModel: ManyToManyLeftSideModel object>' combined with model_attr 'related_models__name' returned None, but doesn't allow a default or null value.

`SearchField.prepare` will now return multiple values when it encounters `ManyToManyFields` somewhere in the lookup chain.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -111,3 +111,4 @@ Thanks to
     * Claude Paroz (@claudep) for Django 1.9 support
     * Chris Brooke (@chrisbrooke) for patching around a backwards-incompatible change in ElasticSearch 2
     * Gilad Beeri (@giladbeeri) for adding retries when updating a backend
+    * Arjen Verstoep (@terr) for a patch that allows attribute lookups through Django ManyToManyField relationships
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -80,39 +80,79 @@ class SearchField(object):
         if self.use_template:
             return self.prepare_template(obj)
         elif self.model_attr is not None:
-            # Check for `__` in the field for looking through the relation.
-            attrs = self.model_attr.split('__')
-            current_object = obj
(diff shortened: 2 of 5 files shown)


Terr:
Travis is reporting a failing test but it seems, looking at other PRs, to be one that is randomly failing.

acdha:
@Terr Yes, sorry about that – there's a race condition with the `multiprocessing` tests which has been hard to reproduce locally. The same test passed after restarting it.

Terr:
No worries, race conditions are always a pain. Thanks for the quick merge!

acdha:
Thanks for the clean patch & tests!

janwin:
this patch breaks our index for foreignKey integerFields with null values allowed

`if len(values) == 1:`
`    return values[0]`
`else:`
`    return values`

should be:

`if len(values) == 1:`
`    return values[0]`
`elif len(values) > 1:`
`    return values`

a null value returns an empty list, which then can not be converted into an integer and causes a type error:

>   File "/usr/local/lib/python3.5/dist-packages/haystack/indexes.py", line 199, in prepare
    self.prepared_data[field.index_fieldname] = field.prepare(obj)
  File "/usr/local/lib/python3.5/dist-packages/haystack/fields.py", line 272, in prepare
    return self.convert(super(IntegerField, self).prepare(obj))
  File "/usr/local/lib/python3.5/dist-packages/haystack/fields.py", line 278, in convert
    return int(value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'

PR #1363, opened 2016-05-16
MultiValueField: fix issues with single value

Terr:
Hi again,

When working with the changes from my [earlier patch](https://github.com/django-haystack/django-haystack/commit/2a450d81432a40c6f0f963b41295f00d7efb5fc7), I noticed that `MultiValueFields` that prepared a single string value ended up as `['S', 't', 'r', 'i', 'n', 'g']` in the search backend. Also, single `int` values (and probably others) resulted in an error. So it seems that I introduced a regression. Apologies. 

I fixed it by battling the symptom of how `MultiValueField.convert` handles the incoming value. I don't think any other fields were affected because they all expect a single value.

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -411,7 +411,10 @@ class MultiValueField(SearchField):
         if value is None:
             return None
 
-        return list(value)
+        if hasattr(value, '__iter__'):
+            return value
+
+        return [value]
 
 
 class FacetField(SearchField):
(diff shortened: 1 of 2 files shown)


Terr:
It seems from Travis that Python 3 returns with some errors. I will dive into it and reopen the PR when I found the issue.

claudep:
We had that `hasattr` issue when migrating Django to Python 3, we have that idiom at several places:
```
if hasattr(value, '__iter__') and not isinstance(value, six.text_type):
```

Terr:
Thanks, that was what I needed!

I have opened a new PR

PR #1364, opened 2016-05-16
Fix MultiValueField issues with single value

Terr:
From my [previous PR](https://github.com/django-haystack/django-haystack/pull/1363):

> Hi again,
> 
> When working with the changes from my earlier patch, I noticed that MultiValueFields that prepared a single string value ended up as ['S', 't', 'r', 'i', 'n', 'g'] in the search backend. Also, single int values (and probably others) resulted in an error. So it seems that I introduced a regression. Apologies.
> 
> I fixed it by battling the symptom of how MultiValueField.convert handles the incoming value. I don't think any other fields were affected because they all expect a single value.

The tests should now also pass in Python 3

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -411,7 +411,10 @@ class MultiValueField(SearchField):
         if value is None:
             return None
 
-        return list(value)
+        if hasattr(value, '__iter__') and not isinstance(value, six.text_type):
+            return value
+
+        return [value]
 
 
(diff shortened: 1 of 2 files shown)


acdha:
Tests look good – we had a failure due to a package download on Travis, which seems to happen sporadically when their systems are under load. Retrying appears to have fixed that.

Terr:
Thanks again :)

PR #1369, opened 2016-05-20
make empty lists return null or default

janwin:
This fix resolves the issue with integerFields trying to convert empty list in case a Nnull value is returned from the datasource

Tests are passing

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -87,7 +87,7 @@ class SearchField(object):
 
             if len(values) == 1:
                 return values[0]
-            else:
+            elif len(values) > 1:
                 return values
 
         if self.has_default():
--- a/test_haystack/test_fields.py
+++ b/test_haystack/test_fields.py
@@ -85,6 +85,14 @@ class SearchFieldTestCase(TestCase):
 
         self.assertEqual([1, 1], field.resolve_attributes_lookup([obj], ['related', 'related', 'value']))
 
+    def test_prepare_with_null_django_onetomany_rel(self):
+        left_model = OneToManyLeftSideModel.objects.create()
+
+        field = SearchField(model_attr='right_side__pk', null=True)
+        result = field.prepare(left_model)
+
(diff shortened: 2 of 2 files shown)


Terr:
Thanks for fixing this. Would you mind if I write some extra tests for this case and open a PR towards your branch?

janwin:
Sure, go on! I would appreciate it!

Terr:
I have some difficulty in reproducing the issue. Am I correct that your setup looks like this:

```
class LeftModel(models.Model):
    some_foreign_key = models.ForeignKey('RightModel', null=True)

class RightModel(models.Model):
    some_integer_field = models.IntegerField()
```

And that your `model_attr` is referring to `some_foreign_key__some_integer_field` ?

Terr:
I think I got it: the ForeignKey relation/column is defined in the 'right' model? I added a test for it [here](https://github.com/Terr/django-haystack/commit/45293cafbed0ef6aeb145ce55573eb32b1e4981f): it fails without your fix and passes with it. 

For some reason I don't see your fork when trying to create a PR, but if you want you can cherry pick the commit.

janwin:
thanks @Terr i added your test to the PR

thongly:
Is this awaiting further review, or tests?

acdha:
@thongly It looks like the CI build was just due to the multiprocessing race condition. If you wanted to rebase this against the current master branch and confirm that the tests pass that'd be helpful

acdha:
The tests just passed after a restart

StevenFerreira:
I've just created a separate pull request minutes ago and all tests passed locally

thongly:
@acdha - Are you planning on pushing a release soon too?  Wondering if we need to pip install the branch-specific version or whether we can wait on your official release.  Please advise if you have a ETA.

acdha:
If you need it right now, I would use a pinned commit. I would like to get
a release fairly soon but didn't have a specific date in mind.

Issue #860, opened 2013-09-20
update_index with multiple workers fails with traceback


comandrei:
Environment: 
  Ubuntu 13.04
  Python 2.7.4
  Solr 4.4.0
  django==1.4.8
  django-haystack==2.0.0
  pysolr==3.1.0

Try to run python manage.py update_index -k 8 and you get a traceback similar to this:

``` python
Traceback (most recent call last):
  File "/home/user/.virtualenvs/ve/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 210, in handle_label
    self.update_backend(label, using)
  File "/home/user/.virtualenvs/ve/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 262, in update_backend
    pool.map(worker, ghetto_queue)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 250, in map
    return self.map_async(func, iterable, chunksize).get()
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 554, in get
    raise self._value
TypeError: cannot concatenate 'str' and 'tuple' objects
```

Running update_index without any parameters works.

viciu:
It is definitely still a problem (!). I experience the same issue when using -k parameter.

I experience it on mysql and elasticsearch 1.6.0

django-haystack==2.4.0

```
Traceback (most recent call last):
  File "/home/user/.virtualenvs/project/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 188, in handle_label
    self.update_backend(label, using)
  File "/home/user/.virtualenvs/project/local/lib/python2.7/site-packages/haystack/management/commands/update_index.py", line 239, in update_backend
    pool.map(worker, ghetto_queue)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 251, in map
    return self.map_async(func, iterable, chunksize).get()
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 558, in get
    raise self._value
SerializationError
```

It looks like multiprocessing is splitting data between processes in a way, that causes (randomly?) different errors, like:

* SerializationError
* JSONDecodeError('Unterminated string starting at: line 1 column 64984 (char 64983)',)
* KeyError: 'items'

and more.

Setting batch size to some small amount reproduces this issue quick:

```
./manage.py rebuild_index -b5 -k2
```

vqw:
I am using:
Django 1.8.11
Whoosh 2.7.4
[...]

PR #1376, opened 2016-05-31
fix: multiprocessing-unsafe session handling in update_worker

acdha:
The `update_index` management command does not handle the `multiprocessing` environment safely. On POSIX systems, `multiprocessing` uses `fork()` which means that when called in a context such as the test suite where the connection has already been used, backends like pysolr or ElasticSearch may have a pre-existing open socket connected to the search server which will be shared *across* workers and HTTP requests are interleaved, producing unexpected errors when the client attempts to decode something like an XML fragment as the HTTP status code.

This commit resets the backend connection inside the workers and in testing has been stable across hundreds of runs, unlike the current situation where a single-digit number of runs would almost certainly have at least one failure.

Other improvements:
* Improved sanity checks for indexed documents in management command test suite. This wasn’t actually the cause of the problem above but since I wrote it while tracking down the real problem there’s no reason not to use it.
* update_index now checks that each block dispatched was processed to catch silent failures.

See #1001 and possibly the most recent comment on #860

### DIFF
--- a/haystack/management/commands/update_index.py
+++ b/haystack/management/commands/update_index.py
@@ -88,9 +88,10 @@ def do_update(backend, index, qs, start, end, total, verbosity=1, commit=True,
                                                                              retries + 1,
                                                                              max_retries))
             break
-        except (SystemExit, KeyboardInterrupt):
-            raise
         except Exception as exc:
+            # Catch all exceptions which do not normally trigger a system exit, excluding SystemExit and
+            # KeyboardInterrupt. This avoids needing to import the backend-specific exception subclasses
+            # from pysolr, elasticsearch, whoosh, requests, etc.
             retries += 1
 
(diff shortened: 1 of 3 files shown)


haystack/management/commands/update_index.py
```diff
@@ -84,10 +88,9 @@ def do_update(backend, index, qs, start, end, total, verbosity=1, commit=True,
                                                                              retries + 1,
                                                                              max_retries))
             break
+        except (SystemExit, KeyboardInterrupt):
+            raise
```
claudep: AFAIK those two lines are unneeded, any reason to include them anyway?

acdha: Good point – I reworked that code several times while testing this and that's cruft

PR #1396, opened 2016-07-01
Fixes #768, Make decimal field actually return one

### LINKED ISSUE #768: DecimalField doesn't convert to Decimal type
Hi,

The current implementation of the `DecimalField` doesn't convert to the `decimal.Decimal` type, as it should - Django, for example, does this for its model fields -.

Here's the code for the field, showing that it converts to `unicode` instead:

``` python
class DecimalField(SearchField):
    field_type = 'string'

    def __init__(self, **kwargs):
        if kwargs.get('facet_class') is None:
            kwargs['facet_class'] = FacetDecimalField

        super(DecimalField, self).__init__(**kwargs)

    def prepare(self, obj):
        return self.convert(super(DecimalField, self).prepare(obj))

    def convert(self, value):
        if value is None:
            return None

        return unicode(value)
```

For a matter of coherence, I'd like to ask that it gets changed to convert do `Decimal` instead. What do you think? Would this present a negative impact to any other part of the app, apart from breaking this part of the API?

richardasaurus:
https://github.com/django-haystack/django-haystack/issues/768


haystack/fields.py
```diff
@@ -298,7 +299,7 @@ def convert(self, value):
 
 
 class DecimalField(SearchField):
-    field_type = 'string'
+    field_type = 'float'
```
acdha: I'm thinking that we should set type to `decimal` here so the backends have the option of using a decimal type if possible

richardasaurus: will try to do that when i get some time

PR #1397, opened 2016-07-02
Solr610

biggosh:
solr.xml template has been aligned with solr 6.1.0 requirements


GabrielDumbrava:
Why did you add requirements for Django 1.9.7?

biggosh:
I mistakely pushed also the requirements.txt file.
It's no reason for this requirements: I've just changed the solr.xml template in accordance to what I found and I'm using with solr610.

PR #1400, opened 2016-07-04
Update how facet forms / views operate

SteveByerly:
I was working on updating the generic_views and noticed that the faceting views/forms operated in a way I did not expect. Specifically, it required the view to pass in data from the GET Querydict as a kwarg to the form, then the form saved it as a property.

This update aims to make the code follow a more standardized process, letting the form handle the processing and cleaning of the data, and also making it a bit easier to customize.

### DIFF
--- a/docs/faceting.rst
+++ b/docs/faceting.rst
@@ -254,7 +254,7 @@ might look like this::
                     <dt>Author</dt>
                     {# Provide only the top 5 authors #}
                     {% for author in facets.fields.author|slice:":5" %}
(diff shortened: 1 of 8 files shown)


edwardoliveira:
It would be nice to have this merged as setting form_class as `FacetedModelSearchForm` in a  `FacetedSearchView` will blow up the code.

acdha:
We should get this rebased against master, too

edwardoliveira:
Hey @SteveByerly, are you still willing to resume working on this patch? Maybe I can try to take a stab at this patch. Thanks!

PR #1402, opened 2016-07-05
refactor highlighting. add new option to prevent text from being trimmed

SteveByerly:
I started this update because I wanted a way to prevent the highlighter from always trimming the text the way it does:
`...<span class="highlighted">this</span> wasn't  ideal`

Since I was adding a fourth option, the current initialization didn't seem ideal (checking the kwargs in various places), so I turned to looping over the passed-in values and setting them in an options dict that gets passed to the Highlighter as its kwargs. Updating how the tag is called helped ease this process, and I also think using `key=value` is a bit more readable.

let me know what you think and if you have any changes you'd like to see.

### DIFF
--- a/haystack/utils/highlighting.py
+++ b/haystack/utils/highlighting.py
@@ -6,24 +6,17 @@ from django.utils.html import strip_tags
 
 
 class Highlighter(object):
-    css_class = 'highlighted'
-    html_tag = 'span'
-    max_length = 200
     text_block = ''
 
-    def __init__(self, query, **kwargs):
+    def __init__(self, query, max_length=200, html_tag='span', css_class='highlighted', trim_text=True):
         self.query = query
 
-        if 'max_length' in kwargs:
-            self.max_length = int(kwargs['max_length'])
+        self.max_length = int(max_length)
+        self.html_tag = html_tag
+        self.css_class = css_class
+        self.trim_text = trim_text
 
-        if 'html_tag' in kwargs:
-            self.html_tag = kwargs['html_tag']
-
-        if 'css_class' in kwargs:
-            self.css_class = kwargs['css_class']
-
-        self.query_words = set([word.lower() for word in self.query.split() if not word.startswith('-')])
+        self.query_words = set([word for word in query.lower().split() if word[0] != '-'])
 
     def highlight(self, text_block):
         self.text_block = strip_tags(text_block)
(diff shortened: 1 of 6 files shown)


SteveByerly:
going to try to include exact phrase highlighting.

haystack/utils/highlighting.py
```diff
@@ -157,9 +149,15 @@ def render_html(self, highlight_locations=None, start_offset=None, end_offset=No
         highlighted_chunk += text[matched_so_far:]
 
         if start_offset > 0:
-            highlighted_chunk = '...%s' % highlighted_chunk
+            if self.trim_text:
+                highlighted_chunk = '...%s' % highlighted_chunk
```
acdha: Wondering whether this should be a Unicode ellipsis character (U+2026, `u'\N{HORIZONTAL ELLIPSIS}'`)

acdha:
One thing I was thinking we should have just for posterity's sake are some tests using non-English text (Chinese, Arabic, something with emoji?) as well as what happens when you have text with very long character runs without spaces (e.g. what if someone pastes a URL or DOI into a text field?) 

Unless I'm missing something it looks like we might have a problem with the string slicing and decomposed Unicode characters with code like this:

```
highlighted_chunk = '%s%s' % (highlighted_chunk, self.text_block[end_offset:])
```

This could produce unexpected results if the highlight boundary fell across a combining character:

```
>>> print(u"\N{LATIN SMALL LETTER N}\N{COMBINING TILDE}")
ñ
>>> print(u"\N{LATIN SMALL LETTER N}\N{COMBINING TILDE}"[0])
n
```

Dealing with this is a mess (it'd probably require [regex](https://pypi.python.org/pypi/regex)'s `\X` or the grapheme cluster iterator in [uniseg](https://pypi.python.org/pypi/uniseg)) so I'd want to test whether this can happen in normal operations and change the design to avoid slicing strings like that where possible.

acdha:
I know this is unrelated to this particular change but I was also wondering whether we could convert this to a `simple_tag` so we could reduce the amount of custom code in use. It looked like we could but that was a very cursory glance.

SteveByerly:
I added exact query highlighting and migrated the template tag to simple_tag. There's also a new `min_query_length` option to limit highlighting of short queries like 'a' or 'is'.

I made a very simple test case for the unicode characters and I'm not sure how to produce a test that would create an issue like you bring up - do you want to shoot me an example?

kopf:
The `test_highlight` test is failing on this branch - I presume this is the only blocker to getting this merged? I'm interested in seeing these changes make their way into trunk. If I help at all, shout.

rez0n:
It is very strange that this has not been accepted for so many years. Without that PR highlighter absolutely useless.
@SteveByerly thanks for this PR.

PR #1407, opened 2016-07-11
Explicit setting of search indexes

fleischie:
Allow users to set specific search index modules via the `HAYSTACK_SEARCH_INDEX_MODULES` setting attribute.

I took care to add test cases for the use cases I modified. Although I must admit, that I didn't took the time to test out several Python/Django versions and combinations.

### DIFF
--- a/haystack/utils/app_loading.py
+++ b/haystack/utils/app_loading.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from django.apps import apps
+from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
 __all__ = ['haystack_get_models', 'haystack_load_apps']
@@ -13,6 +14,13 @@ MODEL = 'model'
 def haystack_get_app_modules():
     """Return the Python module for each installed app"""
     return [i.module for i in apps.get_app_configs()]
+
+def haystack_get_search_indexes():
+    """Return all modules to collect search indexes in"""
+    app_modules = ['%s.search_indexes' % app_mod.__name__ for app_mod in haystack_get_app_modules()]
+    app_modules.extend(getattr(settings, 'HAYSTACK_SEARCH_INDEX_MODULES', []))
+
+    return app_modules
 
 def haystack_load_apps():
     """Return a list of app labels for all installed applications which have models"""
--- a/haystack/utils/app_loading.py
+++ b/haystack/utils/app_loading.py
@@ -17,8 +17,9 @@ def haystack_get_app_modules():
 
 def haystack_get_search_indexes():
     """Return all modules to collect search indexes in"""
-    app_modules = ['%s.search_indexes' % app_mod.__name__ for app_mod in haystack_get_app_modules()]
-    app_modules.extend(getattr(settings, 'HAYSTACK_SEARCH_INDEX_MODULES', []))
+    app_modules = [(app_mod, '%s.search_indexes' % app_mod.__name__) for app_mod in haystack_get_app_modules()]
+    custom_search_indexes = getattr(settings, 'HAYSTACK_SEARCH_INDEX_MODULES', [])
+    app_modules.extend([(None, search_index) for search_index in custom_search_indexes])
 
     return app_modules
 
--- a/haystack/utils/app_loading.py
+++ b/haystack/utils/app_loading.py
@@ -2,7 +2,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from django.apps import apps
-from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
 __all__ = ['haystack_get_models', 'haystack_load_apps']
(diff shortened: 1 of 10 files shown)


SteveByerly:
Thanks for the PR @fleischie  - Do you want to try to handle the failed builds travis is reporting?

fleischie:
@SteveByerly Yes, sorry. I try to make this work. :)

fleischie:
Sorry, this took so long. :(
But I'm glad the build is finally successful. :)

haystack/utils/app_loading.py
```diff
@@ -14,6 +15,14 @@ def haystack_get_app_modules():
     """Return the Python module for each installed app"""
     return [i.module for i in apps.get_app_configs()]
 
+def haystack_get_search_indexes():
```
SteveByerly: I think I would move this to utils/loading.py since it's not really loading apps per-se. It would also keep the logic closer to where it's used.

haystack/utils/app_loading.py
```diff
@@ -14,6 +15,14 @@ def haystack_get_app_modules():
...
+    """Return all modules to collect search indexes in"""
+    app_modules = [(app_mod, '%s.search_indexes' % app_mod.__name__) for app_mod in haystack_get_app_modules()]
+    custom_search_indexes = getattr(settings, 'HAYSTACK_SEARCH_INDEX_MODULES', [])
+    app_modules.extend([(None, search_index) for search_index in custom_search_indexes])
```
SteveByerly: Would it make more sense to set the app_mod here instead of setting to None? That way a user could set the `HAYSTACK_SEARCH_INDEX_MODULES` to a directory instead of listing out specific files. That would let the `module_has_submodule` check in loading.py handle collecting from multiple modules in the dir.

haystack/utils/loading.py
```diff
@@ -186,19 +186,21 @@ def indexes(self):
...
 
-        for app_mod in haystack_get_app_modules():
+        for module, search_index in haystack_get_search_indexes():
             try:
-                search_index_module = importlib.import_module("%s.search_indexes" % app_mod.__name__)
+                search_index_module = importlib.import_module(search_index)
             except ImportError:
-                if module_has_submodule(app_mod, 'search_indexes'):
+                module_splits = search_index.split('.')
+                if (module is not None and
```
SteveByerly: as mentioned in app_loading, perhaps not set the first item in the tuple to None so we can simply provide an app dir in our setting.

test_haystack/test_loading.py
```diff
@@ -350,3 +350,41 @@ def test_more_advanced_get_facet_field_name(self):
         self.assertEqual(self.ui._facet_fieldnames, {'bare_facet': 'bare_facet', 'title': 'title_facet', 'author': 'author_exact'})
         self.assertEqual(self.ui.get_facet_fieldname('title'), 'title_facet')
         self.assertEqual(self.ui.get_facet_fieldname('bare_facet'), 'bare_facet')
+
+    def test_collect_app_modules_indexes(self):
```
SteveByerly: If these functions get moved from `app_loading` to `loading`, then the tests can move too.

haystack/utils/app_loading.py
fleischie: Hi @SteveByerly thanks for the review, I will look into it this week.

Setting the first item to `None` was necessary to correctly re-raise the ImportError in `collect_indexes`. But to be perfectly honest, now that I think about it, I am completely unsure how the workflow is and should be. 😅 

I'll take some time this week to review the PR and update according to your review. 😄

fleischie:
@SteveByerly Hi.

I pulled your PR on my fork, as well as move the collection of the search indexes to `utils/loading.py`.

I think I still don't understand, what you meant with your comment to replace `None` with `app_mod` in the collection routine: I was under the impression, that `module_has_submodule` only checks, whether the `ImportError` is legit (because this resulted from an erroneous import of the custom/default search index module) and raises if the import should have happened (hence the `raise`).

Additionally I don't know, how `inspect.getMembers` work, but if I understand you correctly, there should be an error handler, that takes care of iterating through the package and create new search indexes for each found module in said package (and possibly sub-packages)?

If I misunderstand you, please tell me and I'll rework my solution. I'm happy to help (if it eventually does). 😁

SteveByerly:
Thanks for the updates and follow-up!

My understanding is that `module_has_submodule` will check if the "search_indexes" is a package vs a module. For example, let's say we have an app in our project called "products", which has a directory called "search_indexes" that contains `new.py`and `used.py` modules. In the existing code, haystack is iterating through all the apps and loading up 'search_indexes' then pulling out all the indexes from that file, UNLESS 'search_indexes' is like I describe in the example, where it will instead pull the indexes from each module in the 'search_indexes' directory.

This could work with the code you've written as well. Say when I define a custom search_index in the setting I refer to a package/directory instead of a specific file. For it to work, though, we can't set the module to `None` in the get_indexes function and skip this step in the `module_has_submodule` process. 

So we need to set the first item in the returned tuple to the module of the specified search_index. You can do this with the importlib.import_module and using the dotted path set in the new setting.

fleischie:
@SteveByerly OK. I think I understand now. I will revisit this part and update my PR. Sorry, that this is taking so long and so much of your attention. 😳 But I guess we are on a good path to solving this. 💪

SteveByerly:
Haha, no problem at all! Thank you for spending your time on it!

fleischie:
@SteveByerly To me it doesn't seem, that the current implementation of `django_haystack` pulls all registered search indexes from each named file/module inside `search_indexes` directories/packages.  (Also I did not find matching tests or setups for this)

The `module_has_submodule` part only checks, if, after an ImportError, the imported module has a `search_indexes` submodule (in the current implementation) and raises if it is (else it returns to the next iteration in the discovered `search_indexes`). The following code only inspects the inlaying classes to import them (the separate ModelIndexes).

I constructed the following cases:

    - test_haystack/
      +-- custom_search_indexes/
          +-- search_indexes/
              +-- bar_index.py
              +-- foo_index.py

    - test_haystack/
      +-- custom_search_indexes/
         +-- custom_search_indexes_package/
             +-- bar_index.py
             +-- foo_index.py

Neither with the default, nor with my implementation was I able to import the indexes in the innermost modules. I am by no means an expert on this matter, but I would suggest this feature (importing arbitrarily deeply nested search modules inside given packages/modules) should be in another pull request, that builds on top this one (if it does work ok).

If there is anything else I can do or modify or help with, just let me know, though. 😃

Issue #1424, opened 2016-08-24
Woosh supports boost since 2.0


DylannCordel:
Hi,

In code [whoosh backend, around line 194](https://github.com/django-haystack/django-haystack/blob/master/haystack/backends/whoosh_backend.py#L194): 

```python
# Document boosts aren't supported in Whoosh 2.5.0+.
if 'boost' in doc:
    del doc['boost']
```

But Whoosh supports document boost (from 2.0 to current version). Check [here (2.0 version)](https://bitbucket.org/mchaput/whoosh/src/c159748f54665733d903ffc071db967392a65d6d/src/whoosh/writing.py?at=default&fileviewer=file-view-default#writing.py-185) or [here for the master version](https://bitbucket.org/mchaput/whoosh/src/e344fb64067e45d47ec62dc65a75a50be51264a7/src/whoosh/writing.py?at=default&fileviewer=file-view-default#writing.py-389)

In whoosh backend, we should have:

```python
if 'boost' in doc:
    doc['_boost'] = doc['boost']
    del doc['boost']
```

acdha:
I'll close this to keep the discussion in #1425

PR #1425, opened 2016-08-24
Add doc boost for woosh backend

DylannCordel:
See django-haystack/django-haystack#1424 for more details


acdha:
This looks good to me. I'm thinking we should have a test which at least uses mock to confirm that the boost is working or at least that the value is passed to the backend.

SteveByerly:
Hey @DylannCordel - just wanted to check if you'd be able to add some tests to check these changes are working for the whoosh backend?

DylannCordel:
Yes, @SteveByerly but I can't say when :(

It's always in my Todolist, with other old PR for django-cms...

PR #1426, opened 2016-08-28
Solr version bump

MHM5000:
For those Copy & Paste guys...
:8ball:


MHM5000:
Actually start.jar is no longer accessible in Solr. 
There should be more editing than this.

acdha:
@MHM5000 This is a perennial source of maintenance overhead. I've been looking into tackling this for testing by using the official Docker containers, which might be a good alternative alternative for this as well if you have time to work on this: https://github.com/django-haystack/pysolr/compare/master...acdha:dockerized-test-solr

Issue #1416, opened 2016-07-28
spelling_suggestion with solr can't handle multiple word queries


do3cc:
This time I didn't test if it would work with solr 5, but for my last issue I checked if there were mayor changes in Solr regarding spelling component, and there wasn't So I assume that the problem I observed with Solr 6 is also valid with Solr 5.

When I query with two words, solrs gives me results for both words, independently. Here is what haystack gets when it hoped to receive a single string:
{'endOffset': 10,
 'numFound': 1,
 'origFreq': 0,
 'startOffset': 5,
 'suggestion': [{'freq': 22, 'word': 'peak'}]}

This is not a simple bug.
Either spelling_suggestion tries to build a new single query, working on the suggestions, or the spelling_behavior method changes it response type to return more data.
Personally for now I'd like to at least have the raw spelling information in an internal data structure.

do3cc:
This is related to #1415

Issue #1415, opened 2016-07-28
Spellchecking triggers an AssertionError With solr 5.5 and 6.1


do3cc:
* [x] Tested with the latest Haystack release
* [x] Tested with the current Haystack master branch

## Expected behaviour
I do a query and call spellcheck_suggestion() which then returns a spellcheck suggestion

## Actual behaviour
I get an Assertion error:
```
  File "/usr/local/lib/python3.4/site-packages/haystack/backends/solr_backend.py", line 389, in _process_results
    assert spelling_suggestion is None or isinstance(spelling_suggestion, six.string_types)
```
In the browser I can see that spelling_suggestion is this:
```
spelling_suggestion 	: {'collationQuery': 'text:(highli)',
                                'hits': 4,
                                'misspellingsAndCorrections': ['highligh', 'highli']}
```
In Solr I have this data from spelling suggestion:
```
  "spellcheck":{
    "suggestions":[
      "highligh",{
        "numFound":1,
        "startOffset":5,
        "endOffset":13,
        "origFreq":0,
        "suggestion":[{
            "word":"highli",
            "freq":4}]}],
    "correctlySpelled":false,
    "collations":[
      "collation",{
        "collationQuery":"text:highli",
        "hits":4,
        "misspellingsAndCorrections":[
          "highligh","highli"]}]}}

```
My request handler looks like this:
```
   <requestHandler name="/select" class="solr.SearchHandler">
     <lst name="defaults">
       <str name="echoParams">explicit</str>
       <int name="rows">10</int>
       <str name="spellcheck.dictionary">default</str>
       <str name="spellcheck">on</str>
       <str name="spellcheck.extendedResults">true</str>
       <str name="spellcheck.count">10</str>
       <str name="spellcheck.alternativeTermCount">5</str>
       <str name="spellcheck.maxResultsForSuggest">5</str>
       <str name="spellcheck.collate">true</str>
       <str name="spellcheck.collateExtendedResults">true</str>
       <str name="spellcheck.maxCollationTries">10</str>
       <str name="spellcheck.maxCollations">5</str>
       <!-- <str name="df">text</str> -->
     </lst>
     <arr name="last-components">
       <str>spellcheck</str>
     </arr>
</requestHandler>
```
My spellchecker configuration looks like this:
```
    <lst name="spellchecker">
      <str name="name">default</str>
      <str name="field">text</str>
      <str name="classname">solr.DirectSolrSpellChecker</str>
[...]

PR #1433, opened 2016-09-05
Solr 5+ JSON response support for spell checks

SiyuanQi:
Solr 5+ changed the JSON response format so the suggestions of spell check will be key-value mapped rather than simply paired elements in a list. Now it works for both the default Solr setting and "spellcheck.collateExtendedResults = true".

### DIFF
--- a/haystack/backends/solr_backend.py
+++ b/haystack/backends/solr_backend.py
@@ -392,6 +392,14 @@ class SolrSearchBackend(BaseSearchBackend):
             elif len(raw_results.spellcheck.get('suggestions', [])):
                 spelling_suggestion = raw_results.spellcheck['suggestions'][-1]
 
+            if isinstance(spelling_suggestion, dict):
+                # Solr 5+ JSON response format
+                if isinstance(spelling_suggestion.get('suggestion', [None])[-1], dict):
+                    # Solr setting: spellcheck.extendedResults = true
+                    spelling_suggestion = spelling_suggestion['suggestion'][-1]['word']
+                elif isinstance(spelling_suggestion.get('suggestion', [None])[-1], str):
+                    spelling_suggestion = spelling_suggestion['suggestion'][-1]
+
             assert spelling_suggestion is None or isinstance(spelling_suggestion, six.string_types)
 
         unified_index = connections[self.connection_alias].get_unified_index()
--- a/haystack/backends/solr_backend.py
+++ b/haystack/backends/solr_backend.py
@@ -193,7 +193,7 @@ class SolrSearchBackend(BaseSearchBackend):
                 # this makes option dicts shorter: {'maxAnalyzedChars': 42}
                 # and lets some of options be used as keyword arguments: `.highlight(preserveMulti=False)`
                 kwargs.update({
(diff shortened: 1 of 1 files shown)


SiyuanQi:
This commit is related to #1415, but has not taken care of collations yet. Could be improved to handle collations also.

SteveByerly:
Thank you very much for the PR to address this issue. I spent some time looking into the original issues (#1415 and #1416) and added some additional findings/comments. I think your PR is on the right track, but I'd like to get some more feedback on the original issues before including these changes.

In the meantime, can you please include test cases to test differences in the values returned from the backend?

haystack/backends/solr_backend.py
```diff
@@ -392,6 +392,14 @@ def _process_results(self, raw_results, highlight=False, result_class=None, dist
             elif len(raw_results.spellcheck.get('suggestions', [])):
                 spelling_suggestion = raw_results.spellcheck['suggestions'][-1]
 
+            if isinstance(spelling_suggestion, dict):
+                # Solr 5+ JSON response format
+                if isinstance(spelling_suggestion.get('suggestion', [None])[-1], dict):
```
SteveByerly: you might be able to clean this up a bit by pulling the data out once into a new var, or overriding `spelling_suggestion` and using that.

foo = spelling_suggestion.get('suggestion', [None])[-1]
if isinstance(foo, dict):
 ...

SiyuanQi:
I just improved the commit a bit.
The following are two test cases for ```spellcheck.extendedResults = true``` and ```spellcheck.extendedResults = false```.

#### Test case 1:
Solr setting: ```spellcheck.extendedResults = true```
query: tast
url: http://localhost:8983/solr/eigenTunes/select?indent=on&q=tast&wt=json
``` json
{
  "responseHeader":{
    "status":0,
    "QTime":1133,
    "params":{
      "q":"tast",
      "indent":"on",
      "wt":"json",
      "_":"1473250982209"}},
  "response":{"numFound":0,"start":0,"docs":[]
  },
  "spellcheck":{
    "suggestions":[
      "tast",{
        "numFound":2,
        "startOffset":0,
        "endOffset":4,
        "origFreq":0,
        "suggestion":[{
            "word":"test",
            "freq":3},
          {
            "word":"test0",
            "freq":1}]}],
    "correctlySpelled":false,
    "collations":[
      "collation","test",
      "collation","test0"]}}
```

#### Test case 2:
Solr setting: ```spellcheck.extendedResults = false```
query: tast
url: http://localhost:8983/solr/eigenTunes/select?indent=on&q=tast&wt=json
``` json
{
  "responseHeader":{
    "status":0,
    "QTime":0,
    "params":{
      "q":"tast",
      "indent":"on",
      "wt":"json",
      "_":"1473371615865"}},
  "response":{"numFound":0,"start":0,"docs":[]
  },
  "spellcheck":{
    "suggestions":[
      "tast",{
        "numFound":2,
        "startOffset":0,
        "endOffset":4,
        "suggestion":["test",
          "test0"]}],
    "collations":[
      "collation","test",
      "collation","test0"]}}
```

Issue #1411, opened 2016-07-13
Nested model_attr lookup doesn't work as expected with iterable fields


wylee:
2a450d8 changed how nested model_attr lookups such that if a model field is iterable, the nested lookup may fail.

* [x] Tested with the latest Haystack release
* [x] Tested with the current Haystack master branch

## Expected behaviour

Indexing a nested attribute of an iterable field should work.

## Actual behaviour

Indexing a nested attribute of an iterable field does _not_ work in some cases, specifically with GeoDjango `PointField`s (and probably with any field type that is iterable).

## Steps to reproduce the behaviour

1. Create a model with a GeoDjango `GeometryField` (e.g., `geom = models.PointField()`)
2. Create a `SearchIndex` for that model with a field like `longitude = indexes.FloatField(model_attr='geom__x', indexed=False)`
3. Index the model using `manage.py rebuild_index`
4. `rebuild_index` fails with a message like "haystack.exceptions.SearchFieldError: The model '-122.67912' does not have a model_attr 'x'." (this happens because `SearchField.resolve_attributes_lookup()` iterates over the `x` and `y` fields of GeoDjango `Point`s)

## Configuration

* Operating system version: N/A
* Search engine version: N/A
* Python version: 3
* Django version: 1.8.13
* Haystack version: 2.5.0

--

I don't understand what 2a450d8 meant to do beyond supporting many-to-many fields, so I'm not sure what a proper fix would be, but here's a hacky fix that works for my specific use case (but causes some of the tests introduced in 2a450d8 to fail):

```diff
diff --git a/haystack/fields.py b/haystack/fields.py
index 0fa1abc..25adb54 100644
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -153,7 +153,7 @@ class SearchField(object):
                 return current_objects.all()
             return []

-        elif not hasattr(current_objects, '__iter__'):
+        else:
             current_objects = [current_objects]

         return current_objects
```

SteveByerly:
@wylee I'm not familiar with using GeoDjango, but from the little bit I was able to research, I was unable to see where `x` and `y` are defined. Does it come straight from the DB?

Ultimately, i believe the main issue is that `x` on `geom` has an __iter__ method. If it were a normal string or float or int, etc, it would work like any other attribute lookup. From your example, I would expect the value the index uses to simply be '-122.67912'.

[...]

PR #1441, opened 2016-09-28
Fix object not iterable

henriquechehad:
Fix of "'Model' object is not iterable", occurs when save a object and update the index.

captura de tela 2016-09-28 as 12 36 39

captura de tela 2016-09-28 as 12 36 22

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ env3
 # Build artifacts from test setup
 *.tgz
 test_haystack/solr_tests/server/solr4/
+
+haystack_tests.db
--- a/AUTHORS
+++ b/AUTHORS
@@ -114,3 +114,4 @@ Thanks to
     * Arjen Verstoep (@terr) for a patch that allows attribute lookups through Django ManyToManyField relationships
     * Tim Babych (@tymofij) for enabling backend-specific parameters in ``.highlight()``
     * Antony Raj (@antonyr) for adding endswith input type and fixing contains input type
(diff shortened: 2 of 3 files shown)


SteveByerly:
Thank you for the PR @henriquechehad 
Can you elaborate on what this change is trying to fix? From the error, I believe there may be an issue with your index as opposed to the code itself. Would you please provide the code for the index that generates the error you're seeing?

henriquechehad:
@SteveByerly the error occurs any of this fields:

```
    user_name = indexes.CharField(model_attr='user__name')
    user_id = indexes.IntegerField(model_attr='user__id')
    user_thumbnail = indexes.CharField(model_attr='user__get_thumbnail')
    user_url = indexes.CharField(model_attr='user__get_absolute_url')
```

Any `model_attr` attr with `model__*`

Documented here: http://django-haystack.readthedocs.io/en/v2.5.0/searchfield_api.html?highlight=model_attr#model-attr

When I run `update_index` or `rebuild_index` it works, but save the object and the runtime processor (`haystack.signals.RealtimeSignalProcessor`) try to update the index shows the error: `'User' object is not iterable` or any of other model.

henriquechehad:
Seems is the same problem here: https://github.com/django-haystack/django-haystack/issues/1411

henriquechehad:
Created an issue about the tests failed: https://github.com/django-haystack/django-haystack/issues/1442

Issue #1437, opened 2016-09-19
Django 1.10 release


ellmetha:
Hi!

I wonder: is it planned to release a version of *django-haystack* that is compatible with Django 1.10? I noticed that Travis builds pass for ``Django==1.10a1``. However the setup module of *django-haystack* prevents the use of Django 1.10 (https://github.com/django-haystack/django-haystack/blob/master/setup.py#L16).

Thanks!

acdha:
It's planned but that's in setup.py because we need some help to get it finished. #1434 had much of it working but that didn't reach the point of the test suite completely passing. If you have time to work on this, I would start there.

pa-jama:
+1, in fact I'm currently still using django 1.8 because haystack 2.5 hasn't been released yet which supports 1.9

acdha:
2.5 was released:

https://pypi.python.org/pypi/django-haystack/2.5.0

We definitely need to get a 1.10 support release out soon, however.

derekolsson:
Any news on this?

acdha:
@derekolsson Nothing yet but if you have time to work on this, I'm sure your contributions would be appreciated by the community

patrys:
Is there an issue tracking what remains to be done for 1.10 compatibility? We can consider donating the necessary man power but need to know the scope of work first.

acdha:
@patrys I think #1434 was the furthest along - it was down to a few failing tests but I haven't looked to see how involved the changes appear to be.

SalahAdDin:
:+1: For this!!!

acdha:
For anyone interested in this, if you have time to test the code in #1445 with your apps please update the ticket with the results

artursmet:
I tested this branch with my project, also with a search queryset that contained `.only()` queries. It works, tested with elasticsearch backend.

derekolsson:
Tested it as well in my project with all sorts of queries on elasticsearch backend, haven't seen any issues so far either. Thanks so much for this!

acdha:
I just tagged v2.5.1

PR #1434, opened 2016-09-06
Add official support for Django 1.10

claudep:

PR #1445, opened 2016-10-19
Add support for Django 1.10

ellmetha:
*Support for Django 1.10*
Refs: #1437, #1434

I had to fix the ``haystack.utils.default_get_identifier`` function so that it returns a correct identifier when used with instances whose fields have been deferred. Django 1.10 removed the dynamic model classes that were generated when using ``QuerySet.defer()`` and ``only()``: some tests were failing because of that (the ones related to the use of deferred objects with MLT queries).


acdha:
Tagged v2.5.1 – thanks again!

Issue #1117, opened 2014-12-02
Bad Assumption in SearchQuerySet._fill_cache Leads to Performance Degradation


speedplane:
The `SearchQuerySet._fil_cache` function currently initializes the cache with a large number of `None` entries:

```
        # Setup the full cache now that we know how many results there are.
        # We need the ``None``s as placeholders to know what parts of the
        # cache we have/haven't filled.
        # Using ``None`` like this takes up very little memory. In testing,
        # an array of 100,000 ``None``s consumed less than .5 Mb, which ought
        # to be an acceptable loss for consistent and more efficient caching.
        if len(self._result_cache) == 0:
            self._result_cache = [None for i in range(self.query.get_count())]
```

The assumption that this does not incur a performance penalty is incorrect. I have about 4M documents in my index and this single line takes about 800ms per call, the better part of my entire web request. I'm sure it slows down other code too because it needs so much memory, everything else gets kicked out of cache/ram.

Rather than using a standard list for the `_result_cache`, we should be using a sparse list.  One implementation (which appears relatively fully featured) is here: https://pypi.python.org/pypi/sparse_list

speedplane:
The CPU performance of this issue has been fixed, but the RAM usage has not. Haystack is running on my site which has over [100M results](https://www.docketalarm.com/search/?q=). With this many results, this little line consumes 500MB of RAM! Should be fixed.

acdha:
Agreed. I took a quick look (well, not so quick after discovering that most Python memory profiling tools haven't been updated to support Python 3) at sparse_list and it looks roughly like what you'd expect: a small size increase for the shortest lists and a big win as the data sizes get larger:

https://gist.github.com/acdha/030bccba2c751813dfb0daa5e5344a7a

I'm going to test this later tonight – I need to go vote first – but it looks like it'll be a dependency plus changing that line to be roughly `self._result_cache = SparseList(self.query.get_count())` and then testing. (One potential gotcha: SparseList doesn't have a `__setslice__` – not sure whether we use that or how much it really matters)

One thing I wanted to test was performance dealing with paging through a query set – my quick test above is definitely not representative of more realistic access patterns.

acdha:
[...]

PR #1451, opened 2016-11-09
Use sparse_list to save memory (see #1117)

acdha:
The sparse list implementation avoids allocating the result cache elements
until something triggers loading the actual content.

This has a slight memory increase for very small querysets – roughly a custom
class + dict() vs. a list() – but the actual result data will usually be much
larger and the difference is cancelled out by the time the result cache is at a
thousand items.

This branch is currently broken using the released version of sparse_list – see
https://github.com/johnsyweb/python_sparse_list/pull/5


haystack/query.py
```diff
@@ -233,7 +234,7 @@ def _fill_cache(self, start, end, **kwargs):
         # an array of 100,000 ``None``s consumed less than .5 Mb, which ought
```
speedplane: I would change this comment here explaining that a sparse list is used because the total results set can be many orders of magnitude larger than the set of results that we pull at any given time.

PR #1458, opened 2016-12-02
reStructuredText link format fixes.

jheasly:
Doh! My first attempt at a `reStructuredText` doc. Didn't understand inline links formatting, at all. Now I sorta do.


acdha:
Thanks – the web UI could use a ReST preview mode 😀

jheasly:
@acdha Yeah, no kidding. So I could preview my fix, I went and installed an Atom ReST preview plugin (that required a build of ghc that took _forever_!).

PR #1459, opened 2016-12-05
Fix "more like this" w/ filter

divergentdave:
This is a follow-up from #629 by @JoshData. I pulled out the more_like_this change and added a test to cover it. Previously, if you took `more_like_this()` results and called `filter()` on it, the result would "forget" about the more_like_this part of the query, and you would get all documents that match the filter. (in primary key order) The new test does just this, and then checks the results. The fix is straigthforward, updating `_clone()` to copy both `_more_like_this` and `_mlt_instance`.


JoshData:
Thanks!

PR #1464, opened 2016-12-22
test to test doc boost and field weight

jwnorman:
I made a test to test Whoosh's _boost using @DylannCordel's PR https://github.com/django-haystack/django-haystack/pull/1425. Can you take a look? If this is sufficient, it'd be nice to get this into 2.5.2 or whatever the next version is so that boost will be "reprecated". :)


jwnorman:
@acdha (or others), do you know if you'll have time to take a look at this anytime soon?

jwnorman:
@acdha would it be possible to get an idea of when this might be reviewed? i totally understand you (we all) are busy doing other things.

PR #1484, opened 2017-02-15
update_index --age option can be hours or minutes.

hornn:
Format: `<number>` or `<number>h` - granularity of hours.
`<number>m` - granularity of minutes.


hornn:
There is a workaround/alternative to using `--age`, by using the `--start` option, and modifying it to be `now() - <required increment>`. 
However, this approach is error prone because `now()` can be either django's timezone or the generic datetime, and it will be better to handle this logic consistently inside the update_index code, where `age` is converted to `start_time`.
In addition, when calling it from the command line it's harder to calculate the start time in the same way, and using age is much easier and clearer.

PR #1489, opened 2017-03-14
Removed potential duplicate argument in method call

anibalpacheco:


acdha:
Do you have an example of why this is necessary?

PR #1495, opened 2017-03-30
Fixes in more_like_this function

barseghyanartur:
Seems like this hasn't yet been used or tested. Fix works for me.


acdha:
Can you fix the tests on this so it passes?

PR #1496, opened 2017-04-06
Fix django template context passing

RabidCicada:
Just wanted to fix build_solr_schema failing for django 1.10


RabidCicada:
Reworking...didn't realise there was a test that expected context to be a context object

Issue #1497, opened 2017-04-07
FEATURE: Add support for UUID Primary Keys


RabidCicada:
Currently, haystack fails on search results when the primary key of a model is a UUID.  This is because the INDENTIFIER_REGEX used in get_haystack_identifier()  is too stringent for them in haystack/utils/__init__.py.  It can be worked around by providing your own get_haystack_identifier(), but native support seems to make sense especially as haystack intends to be used in large scale deployments where UUID's are the primary collision avoidance and co-operative ID generation relief.

Insertion works, because it is constructed manually token by token but the retrieval fails.

I'm trying to be a good citizen and have a pull request to fix it all but I'm not sure what "Thing" I should be testing in test_haystack to properly exercise the functionality.

RabidCicada:
Pull request #1498 has been created to solve this issue.  I'm trying to connect the pull request...but I can't see a way to do it.

PR #1498, opened 2017-04-07
Added support for processing uuid primary keys in django. Attempt to solve Issue #1497

RabidCicada:


RabidCicada:
I'm trying to figure out how to connect this pull request to the issue.....but I'm having trouble...any help?

RabidCicada:
I've added a MockModel with UUID for primary key.  I've also added a search index mapping that model.  I don't know where to go from here...what test to throw it in etc....Some guidance here would help me wrap up this PR in a clean and complete manner

RabidCicada:
@Fak3 You want me to consume your changes in PR #1515?  I can add the \w part to my changesets regex and completely replicate your PR I believe.  Plus I've got a head start on the testing:).  That work?

RabidCicada:
BTW...Thanks for reviewing!

PR #1499, opened 2017-04-07
Fix Context object usage in template rendering. Fixes build_solr_schema. Django 1.11

RabidCicada:
Fix a break of build_solr_schema with django 1.11 where they drop support for the Context object.  Also updated test.


RabidCicada:
This also appears to make all tests pass when using django 1.11 in general.

RabidCicada:
This is OBE by the Solr6 upgrade also in pipelines

PR #1503, opened 2017-04-18
WIP: Updated spellcheck response format support

acdha:
This is based on the large Solr 6 overhaul which Kyle Stapp (@RabidCicada) has been undertaking:

https://github.com/RabidCicada/django-haystack/commit/b356097a55d463f62ba00c0f9dc3a5650cc969d8

Since Solr's spelling suggestion response format has changed multiple times over the years, this bit of code deserves more than average levels of testing across recent versions.

This code sets `spellcheck.collateExtendedResults=true`, which requires Solr 4+ but prepares us to allow returning more than a single suggestion.

### DIFF
--- a/haystack/backends/solr_backend.py
+++ b/haystack/backends/solr_backend.py
@@ -204,6 +204,7 @@ class SolrSearchBackend(BaseSearchBackend):
         if self.include_spelling is True:
             kwargs['spellcheck'] = 'true'
             kwargs['spellcheck.collate'] = 'true'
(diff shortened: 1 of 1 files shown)


acdha:
@rabidcicada here's roughly what I was thinking for restructuring this. I'm a little hesitant to enable `spellcheck.collateExtendedResults` but I like the ability to get multiple suggestions.

RabidCicada:
I will review this shortly now that I'll have a little more time.

acdha:
It's probably moot now that the main upgrade branch has most of the same changes

Issue #1502, opened 2017-04-14
Upgrade to 6.X Solr


RabidCicada:
This is a continuation of #1488.

This is for all questions that arise during port.

Or particular interest at the moment:
I'm interested in feedback on upgrading the behavior of spelling suggestions.  Currently it looks like we only return a single result.  I feel this is next to useless other than the collation query.

For the initial port I'm looking to maintain backwards compatibility which looks like returning a SINGLE suggetion with collation results overriding suggestion results (whole query suggestion vs single word suggestion)...but I'm plumbing for future returning all results in a reasonable format.  Does anyone have any input on what data to pass forward?

RabidCicada:
Good news Everyone https://i.ytimg.com/vi/YRCzEqkCoiM/maxresdefault.jpg

All tests passing.

Note of interest: with our default behavior...we never test just the suggestion portion...it always returns a collation.  This is because we pass spellcheck.collate=True with every spellcheck.  This is how it was done previously...but means that we don't EVER test the suggestion spellcheck...only the collate spellcheck.

I plan to change this in the future.  I also would like to return a sensible multiresult in the future ( I preserved what I though was current behavior where we only return one result and only collate if both collate and suggestions are provided)

We now support 6.4.0 and 6.5.0 api wise.

I still need to implement the build_solr_schema...I think later today and the initial port effort is done.
https://github.com/RabidCicada/django-haystack/tree/solr6_template_upgrade

RabidCicada:
I'm currently wrangling with details involving level or responsibility for the command (philosophical crap).  I'm designing the command to take an output 'configure_dir'.  The command will have the same search_configuration/ template directory.  It will also construct the solrconfig.xml from that directory currently feeding the template engine the same context that it feeds the schema.xml.  I'm considering just taking all *.xmls under the search configuration directory and running them through the template engine and dumping them into the configure_dir output directory.

The reason we need solrconfig.xml now is to enable ClassicSchema Factory stuff and disable DynamicFieldHandlers.  In addition there's a defaults field that we should template to play nicely.

This would allow anyone to template in ANY of the config files now using the existing context info (admittedly not very useful outside of schema.xml and solrconfig.xml at the moment -- but future proofing?). 

As it stands I will either:

- Render all xml files through the template engine and into configure_dir then use connections[using].get_backend().reload() to reload the core.

- Only render solrconfig.xml and schema.xml through template engine and into configure_dir then use connections[using].get_backend().reload() to reload the core.

[...]

Issue #1488, opened 2017-03-01
django_ct error with Solr 5.5.4/6.4.1 and Haystack 2.6.0


elitzer2:
* [x] Tested with the latest Haystack release
* [ ] Tested with the current Haystack master branch

## Expected behavior

Haystack responds as expected (with an appropriate JSON response) and is able to be queried using a Solr backend

## Actual behavior

500 Internal Server Error: KeyError: 'django_ct'

## Steps to reproduce the behavior

1. Create Django project and add a model
2. You can try using the `build_solr_schema` command, but that produces an output incompatible (deprecated field types) with both Solr 6.4.1 and Solr 5.5.4 (so I basically have been trying to use [this](https://github.com/nazariyg/Solr-5-for-django-haystack) for the schema. I have tried using the schema output from both cases mentioned in both Solr 6.4.1 and Solr 5.5.4 but have not had any luck.
3. I have ensured the `django_id` and `django_ct` are defined as fields but that does not seem to be the issue here. 

I suppose my main question to the community is whether or not anyone has gotten Haystack working with Solr 5.5.4 or Solr 6.4.1? I don't feel we should have to downgrade Solr any further due to lack of new features and for security reasons. Currently in our Django application we are using PySolr directly to query Solr and return JSON. Thanks for your time!

## Configuration

* Operating system version: Ubuntu 16.04.2 LTS
* Search engine version: Solr 6.4.1 and Solr 5.5.4
* Python version: 3.5.2
* Django version: 1.10.5
* Haystack version: 2.6.0

acdha:
I use Haystack with Solr 6.4.1 in production now but I updated my schema independently since it's fairly customized from what `build_solr_schema` produces. That key error really sounds like Solr doesn't have the expected schema, which makes sense given your point 2 so perhaps the real story here is updating `build_solr_schema` to work with Solr 6.

elitzer2:
If I post my schema file would you mind glancing at it and see if you can find anything obvious? It's only about 90 lines or so and most fieldTypes don't ever get utilized but are just in there from the default schema. Not sure if I should post it directly in here or if I should put it in a Gist or something?

acdha:
A gist would probably be good – the main thing is usually looking at the startup log messages since Solr usually logs the configuration fault but not always in the most obvious form

elitzer2:
Ok I can condense it and get rid of the unused fieldTypes just so it's easier to pinpoint.

elitzer2:
[...]

PR #1315, opened 2016-02-20
Solr 5 support

elishowk:
Continuing and finishing PR #1148 after @acdha comment https://github.com/django-haystack/django-haystack/pull/1148#issuecomment-172620864

Test coverage seems good as well as Travis builds.
Rebase from master is freshly done

Thanks for considering merging it.

---

This change is [Reviewable](https://reviewable.io/reviews/django-haystack/django-haystack/1315)

PR #1504, opened 2017-04-19
Solr6 template upgrade

RabidCicada:
This is the Major upgrade addressing #1502 and started in #1488 to handle Solr6 (specifically 6.4 and 6.5).  It adds new command parameters to build_solr_schema.  The two new options allow you to specify a directory to put the config files in, and to automatically reload the core.  This involved putting a new HAYSTACK_CONNECTIONS variable in place 'ADMIN_URL'.

One major change was that the template got renamed to schema.xml in the template directory instead of solr,xml

Along the way spelling suggestions format parsing was updated and handled (though the legacy style of returning only one suggestion was maintained -- we now have the plumbing to supply multiple).  Upgrades were made to the tests to test spelling suggestion without collation to improve code coverage.

Upgrades were also made to handle django 11 by fixing the context passing to be native dicts in the main code base and the tests.

PEP8 changes were made, though not exhaustively.

Updated the Travis config to only use Trusty images as the others didn't have access to Oracle8 which was also required for newer Solr.
Expanded test matrix to handle Django 1.11

TBD:
Upgrade haystack to handle multiple spelling suggestions for Solr
Something about ExtendedCollation from @acdha's other comments
Expand matrix to handle multiple Solr versions (don't know what a reasonable anchor version is)
probably many more:)


haystack/backends/solr_backend.py
```diff
@@ -389,13 +393,38 @@ def _process_results(self, raw_results, highlight=False, result_class=None, dist
...
-            if len(raw_results.spellcheck.get('collations', [])):
-                spelling_suggestion = raw_results.spellcheck['collations'][-1]
-            elif len(raw_results.spellcheck.get('suggestions', [])):
-                spelling_suggestion = raw_results.spellcheck['suggestions'][-1]
+            # There are many different formats for Legacy, 6.4, and 6.5
+            # e.g. https://issues.apache.org/jira/browse/SOLR-3029
+            collations = raw_results.spellcheck.get('collations', [])
+            suggestions = raw_results.spellcheck.get('suggestions', [])
+            if len(collations):
+                #Handle sol6.5 collation format
```
acdha: Minor nitpick but we should probably have a space after `#` in the comments in this block

haystack/backends/solr_backend.py
```diff
@@ -389,13 +393,38 @@ def _process_results(self, raw_results, highlight=False, result_class=None, dist
...
+                if isinstance(collations, dict):
+                    spelling_suggestions= [col['collationQuery'] for col in collations.values()]  #aggregate for future use in multi suggestion response
+                #Legacy Legacy & 6.4 handling
+                else:
+                    if isinstance(collations[1], dict):  #Solr6.4
+                        spelling_suggestions = [item["collationQuery"] for item in collations if isinstance(item,dict)]  #aggregate for future use in multi suggestion response
+                    else:  #Legacy Solr format
+                        spelling_suggestions=collations[-1]
+
+                spelling_suggestion = spelling_suggestions[-1]  #Keep current method of returning single value
```
acdha: Is there a reason why we would want the last suggestion rather than the first? I'm thinking we should have a comment with the rationale

haystack/backends/solr_backend.py
```diff
@@ -389,13 +393,38 @@ def _process_results(self, raw_results, highlight=False, result_class=None, dist
...
+                else:
+                    spelling_suggestions = []
+                    if isinstance(suggestions[1], dict):  #Solr6.4
+                        for item in suggestions:
+                            if isinstance(item, dict):
+                                spelling_suggestions += [subitem["word"] for subitem in item['suggestion']]
+                    else:  #Legacy Solr
+                        spelling_suggestions=suggestions[-1]
+
+                spelling_suggestion = spelling_suggestions[-1]  #Keep current method of returning single value
```
acdha: Since we have this code in a few places, what do you think about declaring `spelling_suggestions` once, having the various branches populate it, and then having the check to return None or a single value at one place afterwards?

haystack/constants.py
```diff
@@ -10,6 +10,7 @@
 ID = getattr(settings, 'HAYSTACK_ID_FIELD', 'id')
 DJANGO_CT = getattr(settings, 'HAYSTACK_DJANGO_CT_FIELD', 'django_ct')
 DJANGO_ID = getattr(settings, 'HAYSTACK_DJANGO_ID_FIELD', 'django_id')
+HAYSTACK_DOCUMENT_FIELD = getattr(settings, 'HAYSTACK_DOCUMENT_FIELD', 'text')
```
acdha: It seems like this could be `DOCUMENT_FIELD` following the other settings losing the `HAYSTACK_` prefix when retrieved.

haystack/management/commands/build_solr_schema.py
```diff
@@ -22,15 +28,66 @@ def add_arguments(self, parser):
...
+            self.stdout.write("Trying to write schema file located at {}".format(options.get('filename')))
             self.write_file(options.get('filename'), schema_xml)
-        else:
+            if options.get('reload'):
+                connections[using].get_backend().reload()
+
+        if options.get('configure_dir'):
+            cdir = options.get('configure_dir')
+            self.stdout.write("Trying to configure core located at {}".format(cdir))
+            if os.path.isfile(cdir+'/managed-schema'):
```
acdha: We should use `os.path.join` for all of the filesystem path construction logic in this section to avoid issues with Windows

haystack/management/commands/build_solr_schema.py
```diff
@@ -42,17 +99,17 @@ def build_context(self, using):
...
             'default_operator': constants.DEFAULT_OPERATOR,
             'ID': constants.ID,
             'DJANGO_CT': constants.DJANGO_CT,
             'DJANGO_ID': constants.DJANGO_ID,
-        })
+        }
 
-    def build_template(self, using):
-        t = loader.get_template('search_configuration/solr.xml')
+    def build_template(self, using, tfile=schema_template_loc):
```
acdha: Since this is a keyword argument it should be something like `template_filename` which is a little more obvious when someone looks at the help, uses autocomplete in an editor, etc.

haystack/management/commands/build_solr_schema.py
```diff
@@ -3,15 +3,21 @@
...
 from django.template import Context, loader
+from django.conf import settings
 
 from haystack import connections, connection_router, constants
 from haystack.backends.solr_backend import SolrSearchBackend
 
-
+import pysolr
+import os
+import traceback
```
acdha: I use `isort` to sort imports and split them into multiple sections (stdlib, third-party, this project). See 9f8c5833875e27e6e34526e2a3417b970d76e77a

test_haystack/mocks.py
```diff
@@ -56,7 +57,7 @@ def update(self, index, iterable, commit=True):
 
     def remove(self, obj, commit=True):
         global MOCK_INDEX_DATA
-        if commit == True:
+        if commit is True:
```
acdha: Is there a reason why this can't just be `if commit:`?

test_haystack/solr_tests/test_management_commands.py
```diff
@@ -3,20 +3,26 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import datetime
+import pysolr
+import os
 from tempfile import mkdtemp
+try:
+    from StringIO import StringIO
```
acdha: Since we don't support Python 2.6 any more we can make this `from io import StringIO` on everything since Python 2.7 supports that.

test_haystack/solr_tests/test_management_commands.py
```diff
@@ -183,6 +197,54 @@ def test_build_schema_wrong_backend(self):
...
+        needle = 'Th3S3cr3tK3y'
+        constants.HAYSTACK_DOCUMENT_FIELD = needle #Force index to use new key for document_fields
+        settings.HAYSTACK_CONNECTIONS['solr']['URL'] = settings.HAYSTACK_CONNECTIONS['solr']['URL'].rsplit('/',1)[0]+'/mgmnt'
+
+        ui = UnifiedIndex()
+        ui.build(indexes=[SolrMockSecretKeySearchIndex()])
+        connections['solr']._index = ui
+
+        renderedfile = StringIO()
+        orig_dir = get_script_dir()+"/confdir"
```
acdha: We should use `os.path` for these as well

test_haystack/solr_tests/test_management_commands.py
```diff
@@ -183,6 +197,54 @@ def test_build_schema_wrong_backend(self):
...
+        conf_dir = get_script_dir()+"/server/solr/server/solr/mgmnt/conf"
+        schemafile = conf_dir+'/schema.xml'
+        solrcfgfile = conf_dir+'/solrconfig.xml'
+
+        call_command('build_solr_schema',using='solr',stdout=renderedfile)
```
acdha: spaces after commas

test_haystack/solr_tests/test_management_commands.py
```diff
@@ -183,6 +197,54 @@ def test_build_schema_wrong_backend(self):
...
+        with open(schemafile) as s:
+            self.assertTrue(s.read().find("name=\""+needle)!=-1)
+        with open(solrcfgfile) as s:
+            self.assertTrue(s.read().find("name=\"df\">"+needle)!=-1)
+        self.assertTrue(os.path.isfile(conf_dir+'/managed-schema.old'))
+
+        call_command('build_solr_schema',using='solr',reload=True)
+
+        os.rename(schemafile,schemafile+".bak")
+        while os.path.isfile(schemafile):#work around stupid delay in filesystem
```
acdha: What filesystem was this running on that `os.rename` wasn't atomic? If this is necessary, I'd like to at the very least have a timeout so it can't run indefinitely

test_haystack/solr_tests/test_management_commands.py
acdha: n.b. this will require reconciling usage with `io.StringIO` being text (and thus Unicode) vs. binary.

RabidCicada:
You shoudl have commit access to this PR.  I have the checkbox checked for "Allow edits from maintainers."  I would wholeheartedly say make any changes you deem fit.

This is the first time I've been in a collaborative review....So I'll try to get the hang of the review process

test_haystack/solr_tests/test_management_commands.py
RabidCicada: I think that's OBE now.  It was incorrectly diagnosed by me when pysolr.reload was failing to actually reload at all.  I thought it was a race condition. because the files were changing and pysolr wasn't giving me anything bad (because solr was helpfully responding with status on the entire server) as opposed to reloading the core I wanted.

test_haystack/mocks.py
RabidCicada: Happy to shorten.  Just thought a tad more explicit would be fine:).

haystack/backends/solr_backend.py
RabidCicada: This is just the way it was already done.  I have no intuition as to why the last was taken.  I believe solr CAN do some sorting of the returns based on some priority...But just wanted to keep convention.

RabidCicada:
I need help figuring out how to accept your requested changes.  I'm looking all over the interface...can't find out how to "accept"

acdha:
I just pushed my commits to your branch

RabidCicada:
Chris.  I have no problem doing the spelling suggestions the way that you started in your other branch.  I just figure we might as well get in what works currently in this PR then we can massage more.

BTW...Any thoughts on what version this might land in or release timeframe?  I'm excited about not having to custom patch things:).

acdha:
Check out bd877a5ee526a23b0608a2b95c084e340b967fbd – I moved the spelling suggestion parsing code into a function so we can catch exceptions for logging and, if enabled, honor the fail silently backend setting.

I've tested that with Solr 6.1, 6.4 and 6.5, but I was thinking what we should do is collect some JSON responses from each of those (and probably with both `spellcheck.collateExtendedResults` values) in the test suite so we can unit-test the response parsing independently.

RabidCicada:
This weekend I will definitely do some work on the documentation so that we can drop it in.  They way I did it...I feel like we can drop it in and it doesn't regress in any functionality.  So documentation could be done afterward.  But I will be working on docs this weekend.

acdha:
The main thing I was thinking is that this should just be a 2.7.0 release with a clear changelog entry so anyone who has a really old Solr install kicking around knows that they need to review their schema if they're still auto-generating it.

RabidCicada:
Indeed.  It's all backward compatible in terms of schema if they are using their own template EXCEPT that the template file is now expected to be named schema.xml instead of solr.xml

RabidCicada:
Hella sick this weekend.  I will be nibbling at this as I feel better through the week.

acdha:
@RabidCicada No worries – take care of yourself!

test_haystack/solr_tests/test_management_commands.py
acdha: Just waiting on Travis to merge https://github.com/acdha/django-haystack/commit/ebec7e01e7ffde71e9711d766692341d439a3cd9

acdha: n.b. I also renamed this file to `test_solr_management_commands.py` so you don't have to look at the parent directory to realize that it's not the same file as `test_haystack/test_management_commands.py`

acdha:
I think by now the main thing left is figuring out how we can avoid committing the Solr files like the stopwords for each language

RabidCicada:
I'm mid update of the documentation.  I've expanded the mangement_commands help file and kept a minimal solr4 and solr6 portion for "Installing Search Engines".  I crossreference to the command help for build_solr_schema.

I am ok with not checking those stopwords files in.  I just took the shortest path to working core.  I bet we don't need any non-english ones. or maybe any of them at all.  I'm still not a solr expert so I don't know if that somehow affects the results of the solr parsing relevant to our tests.

What do you think about having build_solr_schema render every file in search_configuration through the template system.  This future proofs us against new solr changes in the future where a new file (or any file a user wants) could use the provided fields/id's templated.  I could make a subdir of search_configuration 'search_configuration/solr' specifically for it.  That way we also leave room for anything not solr related (which admittedly is nothing so far).

RabidCicada:
I've taken the shortest path for the PR to finish.  

WRT rendering all files in the template dir: We only render solrconfig.xml and schema.xml.
You can't ask the django system directly for all template files in a directory.  You have to poke around.  I'd be happy to add rendering of all template files int he search_configuration dir in another PR.  I quickly investigated and it became clear that we'd have to add a lot of checks for conventions that changed from pre 1.8 and after 1.8 for django.  TEMPLATE configuration underwent large changes and I've seen different conventions with regard to ROOT_DIR and BASE_DIR etc.

RabidCicada:
Travis is performing maintenance so my second checking failed spuriously....I'll rerun tests in the morning...and you might have to.  Or I just commit something small to force:).

RabidCicada:
Boo...removing the stopwords actually caused problems...I'll look at it tomorrow. Maybe I can config to not need it.  I could just have them create using example stuff then drop our configs in. But then it's possible we'll have a mis-mash of old/new stuff.  Any opinion?  We config completely from scratch (what i was doing with the tons of stopword files) or I change something to not require those stopwords (possible -- but not investigated) or I create from an example and drop solrconfig.xml and schema.xml into place.  Idealy configuring to not need those stop words would be great:).

acdha:
I was hoping to avoid committing anything which we don't maintain and
instead copying them from the unpacked tarball.

RabidCicada:
Alrighty.  I think we are in a good place now.  I removed all but our two per-configured files (solrconfig.xml and schema.xml), and we generate from solr. then overlay our changed files.

RabidCicada:
Let me know if there's anything else.  I believe everything is in order.

RabidCicada:
Eh? :)

haystack/management/commands/build_solr_schema.py
```diff
@@ -2,57 +2,140 @@
...
+
+            solrconfig_path = os.path.join(configure_directory, 'solrconfig.xml')
+
+            try:
+                self.write_file(solrconfig_path, solrcfg_xml)
+            except EnvironmentError as exc:
+                raise CommandError('Could not write {}: {}'.format(solrconfig_path, exc))
+
+        if reload_core:
+            core = settings.HAYSTACK_CONNECTIONS['solr']['URL'].rsplit('/', 1)[-1]
```
ahoho: This line throws an exception if the core is labeled as `'default'` instead of `'solr'` in `HAYSTACK_CONNECTIONS`. Is this the desired behavior?

acdha: That's fixed in def141cf5d08e245fa07331e1115ce647d77298c

haystack/management/commands/build_solr_schema.py
```diff
@@ -94,16 +94,16 @@ def handle(self, **options):
...
 
         if reload_core:
-            core = settings.HAYSTACK_CONNECTIONS['solr']['URL'].rsplit('/', 1)[-1]
+            core = settings.HAYSTACK_CONNECTIONS[using]['URL'].rsplit('/', 1)[-1]
 
-            if 'ADMIN_URL' not in settings.HAYSTACK_CONNECTIONS['solr']:
                 raise ImproperlyConfigured("'ADMIN_URL' must be specified in the HAYSTACK_CONNECTIONS settings for the backend")
-            if 'URL' not in settings.HAYSTACK_CONNECTIONS['solr']:
                 raise ImproperlyConfigured("'URL' to the core must be specified in the HAYSTACK_CONNECTIONS settings for the backend")
+            if 'ADMIN_URL' not in settings.HAYSTACK_CONNECTIONS[using]:
```
ahoho: I don't think I can commit on this branch, so I wanted to point out the minor syntax error here!

acdha: Thanks - that was correct when authored but it looks like I screwed up the merge

acdha: Yeah, just a sloppy split staging those lines. I corrected the history in 03209daf3ec2c05c613b7d8170874e38a0d086df so both commits are valid standalone but the end state of the tree is unchanged

RabidCicada:
Thanks guys for the input and editing.  I just got out from under a giant project so I'll have more free time soon.  Please let me know if there's any more work needed here.

ahoho:
I migrated our project from the fork on #1315 to this one and haven't encountered any issues beyond the minor one I mentioned above (now fixed). I'll be doing more extensive testing this week and next, but I don't think we'll hit any hiccups.

One thing worth noting is that #1315 had a `SolrAdmin` class that used the REST API for schema updates (instead of writing to `/conf/schema.xml`). It was handy, but I don't think it's a necessary feature.

ahoho:
I'm catching some errors that I believe are related to the new `managed-schema` and `solrconfig.xml`. Searches with more complex query syntax return incorrect results, even on the Solr admin page. 

In the past, the following two queries were equivalent
`bodytext:(+jones -goldberg -fred -"casey jones" -kennedy)`
`bodytext:(+jones -(goldberg fred "casey jones" kennedy))`

Now, the second one returns the same number of results as the query `jones`. When I use the Solr admin's query analyzer, I receive errors for both queries (with a blank message). I tried to fool around with logging to see if any errors were thrown, but couldn't figure out how to set up the handlers correctly.

When I use DebugQuery, I'm seeing what might be the problem, but I'm not sure.

In the old version, it returns
```
"rawquerystring":"bodytext:(+jones -(goldberg fred \"casey jones\" kennedy))",
"querystring":"bodytext:(+jones -(goldberg fred \"casey jones\" kennedy))",
"parsedquery":"+bodytext:jone -(bodytext:goldberg bodytext:fred PhraseQuery(bodytext:\"casei jone\") bodytext:kennedi)",
"parsedquery_toString":"+bodytext:jone -(bodytext:goldberg bodytext:fred bodytext:\"casei jone\" bodytext:kennedi)",
```

And the new version returns
```
"rawquerystring":"bodytext:(+jones -(goldberg fred \"casey jones\" kennedy))",
"querystring":"bodytext:(+jones -(goldberg fred \"casey jones\" kennedy))",
"parsedquery":"+bodytext:jone -(+bodytext:goldberg +bodytext:fred +PhraseQuery(bodytext:\"casei jone\") +bodytext:kennedi)",
"parsedquery_toString":"+bodytext:jone -(+bodytext:goldberg +bodytext:fred +bodytext:\"casei jone\" +bodytext:kennedi)",
```

Note the added `+` inside the parenthesis. I'm not familiar enough with Solr to determine what's causing this issue. Thoughts?

haystack/templates/search_configuration/schema.xml
```diff
@@ -0,0 +1,1055 @@
...
+    <uniqueKey>{{ ID }}</uniqueKey>
+
+    <!--
+    Not required if the default search field (<str name="df">text</str>) is defined for
+    /select request handler in sorlconfig.xml
+    -->
+    <defaultSearchField>{{ content_field_name }}</defaultSearchField>
+
+    <!-- SolrQueryParser configuration: defaultOperator="AND|OR" -->
+    <solrQueryParser defaultOperator="{{ default_operator }}"/>
```
ahoho: I think this line is what's causing the problem. Without it, the error I described does not occur.

ahoho:
The other error with the analyzer I mentioned is a red herring, and occurs because some `RequestHanders` were missing:
```xml
  <requestHandler name="/analysis/field"
                  startup="lazy"
                  class="solr.FieldAnalysisRequestHandler" />
  <requestHandler name="/analysis/document"
                  class="solr.DocumentAnalysisRequestHandler"
                  startup="lazy" />
```

haystack/templates/search_configuration/solrconfig.xml
```diff
@@ -0,0 +1,1438 @@
...
+  <requestHandler name="/update/extract"
+                  startup="lazy"
+                  class="solr.extraction.ExtractingRequestHandler" >
+    <lst name="defaults">
+      <str name="lowernames">true</str>
+      <str name="fmap.meta">ignored_</str>
+      <str name="fmap.content">{{ content_field_name }}</str>
+    </lst>
+  </requestHandler>
+
```
ahoho: The analysis tool in the Solr admin also doesn't work because some `RequestHanders` were missing.
```xml
  <requestHandler name="/analysis/field"
                  startup="lazy"
                  class="solr.FieldAnalysisRequestHandler" />
  <requestHandler name="/analysis/document"
                  class="solr.DocumentAnalysisRequestHandler"
                  startup="lazy" />
```

RabidCicada:
I'll look at this today a little bit.  Thanks for the messages.

SalahAdDin:
Good news for `Oscar` and for me!

PR #1508, opened 2017-05-22
Add untokenized SearchField

ahoho:
I've added a new `StringField` class to handle raw strings. There are also faceted and multivalued versions. Unlike `CharField`, these are fields that do not tokenize/analyze strings, but store them in the index as-is.


denizdogan:
Would #1789 solve this too?

acdha:
This definitely looks like a good candidate for rebasing against the current branch and shipping.

PR #1512, opened 2017-05-24
Fix #iss1509

AlvaroLQueiroz:


acdha:
It looks like this will be moot with #1504 which updates `build_solr_schema`. The tests are failing for the other changes because through at least Django 1.11 [`Template.render()`](https://docs.djangoproject.com/en/1.11/ref/templates/api/#django.template.Template.render) still requires a `Context` instance. It looks like the point of confusion was the API change which added pluggable backends which introduced the possibility of `get_template` or `select_template` returning something which is not `django.template.Template`:

https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/#get-template-upgrade-django-18

PR #1515, opened 2017-06-04
utils.default_get_identifier: Support non-numeric object pk

Fak3:
Issue: when non-numeric pk is used on model (UUIDField for ex) it is impossible to remove object from index:
"AttributeError: Provided string 'products.product.1243-4567-1234-6789' is not a valid identifier."


acdha:
Thanks! That looks like an old typo. Can you add some test signatures so we can avoid any future regressions?

https://github.com/django-haystack/django-haystack/blob/176bc54c2c439044ae427e6a30707cf4d74336e8/test_haystack/test_utils.py#L24

Fak3:
@acdha PR #1497 does mostly the same. It has some tests. Could it be merged?

PR #1516, opened 2017-06-06
Use filter instead of narrow for limiting models

Mognom:
Improve search performance by filtering models directly in the query. Currently the whoosh backend is making a query per model (for selecting the available documents) and then it makes the real query filtering the documents using the results from those initial queries.


acdha:
Is this completely functionality-neutral? It looks like the tests are passing but I don't know whether you'd like to review coverage on that.

Mognom:
It should be functionality-neutral, although it doesn't look like narrow queries are being tested throughly, so new tests may need to be added to test it properly.

Issue #1514, opened 2017-06-01
update_index's batched slice re-queries are very slow with large OFFSETs


ggaughan:
As mentioned in issue #1077, running `update_index` against millions of rows (with multiprocessing and lots of database memory) I also saw it get progressively slower with mysql (innodb), only handling a few million after several days.

The `update_index` batching method (whether using multiprocessing or not) gets slower and slower because naive offsets (LIMITs) are used (this is the same issue that cripples almost all paginations of large datasets - see http://use-the-index-luke.com/no-offset).

The loop to split the queryset into batches uses `start:end` slices on a lazy queryset which issues a full re-query and sort of ALL rows (at full-row width) for *every* batch (default = 1000 rows) and then the db-server has to page through the results to find the next offset. For datasets that don't fit into the database cache this will effectively:

 * flush out all other users' data
 * waste i/o sorting the whole queryset
 * waste i/o paging in and out through the leading portion of the results every time, getting progressively slower as the offset increases

I've implemented an alternative approach which:

 * avoids the sort by using the covering index to partition the table
 * avoids the initial count (which I think also makes the partitioning more robust, especially for mysql)
 * partitions the batches into ranges of primary keys that can be located instantly via the index without re-sorting and re-reading any of the batch records

I haven't got round to looking at adding a unit test to show the benefits, but my code is here:

  https://github.com/ggaughan/django-haystack/tree/scalable_update_index

acdha:
This seems like a nice win – if you have time to wrap it into a pull-request, it'd be good to confirm test coverage.

ggaughan:
Well #1517 should do it, but Travis seems stuck...

wetneb:
I think this is stale, offsets no longer seem to be used. However, there is still a `COUNT` query that is used to report the number of objects to update, and even that is impractical in my case. Using the largest pk is much more efficient.

acdha:
@wetneb If you want to send a PR that might be helpful — it's not just the logging line but also the code which chunks the update batches so you might need to rewrite that loop to keep reading blocks until it gets one which is smaller than the block size. I'm thinking that could be a separate issue.

PR #1517, opened 2017-06-07
Scalable update index

ggaughan:
Optimised update_index for partitioning large datasets (for #1514)

### DIFF
--- a/haystack/management/commands/update_index.py
+++ b/haystack/management/commands/update_index.py
@@ -25,11 +25,11 @@ LOG = multiprocessing.log_to_stderr(level=logging.WARNING)
 
 
 def update_worker(args):
-    if len(args) != 10:
+    if len(args) != 9:
         LOG.error('update_worker received incorrect arguments: %r', args)
         raise ValueError('update_worker received incorrect arguments')
 
-    model, start, end, total, using, start_date, end_date, verbosity, commit, max_retries = args
+    model, start, end, using, start_date, end_date, verbosity, commit, max_retries = args
 
     # FIXME: confirm that this is still relevant with modern versions of Django:
     # We need to reset the connections, otherwise the different processes
@@ -57,11 +57,11 @@ def update_worker(args):
     backend = haystack_connections[using].get_backend()
 
     qs = index.build_queryset(start_date=start_date, end_date=end_date)
-    do_update(backend, index, qs, start, end, total, verbosity, commit, max_retries)
+    do_update(backend, index, qs, start, end, verbosity, commit, max_retries)
     return args
 
 
-def do_update(backend, index, qs, start_pk, end_pk, total, verbosity=1, commit=True,
+def do_update(backend, index, qs, start_pk, end_pk, verbosity=1, commit=True,
               max_retries=DEFAULT_MAX_RETRIES):
     current_qs = qs.filter(pk__range=(start_pk, end_pk)).order_by()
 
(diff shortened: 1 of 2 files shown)


wetneb:
@ggaughan Thanks a lot, I had the same problem and ended up coding my own version of update_index in my project to bypass the restrictions of the existing one.

@acdha it would be great if this could be merged!

RabidCicada:
@ggaughan Hey Greg...I haven't viewed the contents of this at all...but if it's finished.  Could you update it to remove the conflicts due to time rot?

Then we can bug @acdha for some more attemtion:)

wetneb:
Thanks @ggaughan for updating the PR! It looks good to go now.

haystack/management/commands/update_index.py
```diff
@@ -57,64 +57,47 @@ def update_worker(args):
...
+def do_update(backend, index, qs, start_pk, end_pk, verbosity=1, commit=True,
+              max_retries=DEFAULT_MAX_RETRIES):
+    current_qs = qs.filter(pk__range=(start_pk, end_pk)).order_by()
 
     is_parent_process = hasattr(os, 'getppid') and os.getpid() == os.getppid()
 
     if verbosity >= 2:
         if is_parent_process:
-            print("  indexed %s - %d of %d." % (start + 1, end, total))
+            print(u"  indexing %s - %s." % (start_pk, end_pk))
```
acdha: Strictly speaking the `u` prefix is unnecessary since we have `from __future__ import unicode_literals` in this file

acdha: While we're in here, I'm also wondering whether we should just switch to logging for anyone who is calling this command from a context other than a normal shell. Maybe that's a ticket for followup work.

haystack/utils/app_loading.py
```diff
@@ -21,7 +21,7 @@ def haystack_load_apps():
 def haystack_get_models(label):
     try:
         app_mod = apps.get_app_config(label)
-        return app_mod.get_models()
+        return sorted(app_mod.get_models(), key=lambda model: u'%s.%s' % (model._meta.app_label, model._meta.object_name))
```
acdha: This feels like this should perhaps have a comment — I'm assuming the intent was just stable ordering across runs? — and I was wondering whether it needs to be a string sort rather than just letting it sort on the values (`key=lambda model: (model._meta.app_label, model._meta.object_name)`).

ggaughan:
Not sure about the non-numeric PK test (there seems to be a MockUUIDModel that could cover it, but it may not be in the tests)

RabidCicada:
Apologies.  The mockuuidmodel is not actually used yet.  I'm going to get it fixed in my next PR to address #1543 .  I have had a lot of real life calamity and chaos this last week so I didn't get to finish it.  Will work on it as soon as real life stuff calms down.

RabidCicada:
I've just created a PR #1551 to address Issue #1543 which also implements UUIDMockModel testing with a variety of tests.  Hopefully this will provide coverage here.  I changed the name of the uuid model to fit naming conventions with the rest of the stuff I see.

RabidCicada:
alright...one more rebase and if tests pass, I think this passes @acdha 's requirements:)

wetneb:
@acdha I think this PR is stale and should be closed. It is super-seeded by https://github.com/django-haystack/django-haystack/pull/1479 which was merged.

PR #1518, opened 2017-06-13
Changed GeoDjango Link in Documentation

mohitkh7:
Changed GeoDjango link from geodjango.org to https://docs.djangoproject.com/en/1.11/ref/contrib/gis/


acdha:
It looks like this was made against a rather old branch. I cherry-picked the update into 93f077d719c93547c70d00b42828271ca4642b01.

PR #1534, opened 2017-07-19
Get whoosh backend working properly with fuzzy search.

december1981:
See comments on related commits.


acdha:
`HAYSTACK_FUZZY_MAX_EXPANSIONS` is only supported by the ES backend currently so perhaps we should make that an option specific to that backend if nothing else supports it.

PR #1535, opened 2017-07-28
Fix haystack.__version__

tomkins:
Minor thing I noticed on some of my sentry bug reports, the version number of the package isn't working.

Before:

```jupyter
In [1]: import haystack

In [2]: haystack.__version__
Out[2]: (0, 0, 'dev0')
```

After:

```
In [1]: import haystack

In [2]: haystack.__version__
Out[2]: '2.6.2.dev76+g8cb00e0'
```


tomkins:
After looking at it again - I've also updated the fallback `__version__` to be a string as well. Otherwise it would be a string if it can be found, and a tuple if not.

acdha:
Re:string / tuple - let's check the upstream source as I thought that had changed relatively recently when I switched to setuptools_scm

bigjust:
@acdha Looks like its currently a tuple 

https://github.com/django-haystack/django-haystack/blob/master/haystack/__init__.py#L18

tomkins:
Updated as it's been a while.

The changes in 7189e786dbf3002bf7d533f904e2395a0abd248f fixed it to be a string, although it still comes up with 0.0.dev0 as `__name__` is `haystack` - but the installed package is django-haystack.

PR #1539, opened 2017-08-17
Add Socialhome to who uses

jaywink:
Socialhome is a decentralized and federated profile builder with social networking features. Haystack is used for searching other users and content. We use Whoosh. The code is at https://github.com/jaywink/socialhome


claudep:
Sorry, but we removed that page from the docs in 51a664069.

PR #1547, opened 2017-08-29
enabled FuzzyTermPlugin to allow fuzzy matching with Whoosh

m-thielen:
The FuzzyTermPlugin is needed to allow fuzzy matching with Whoosh (2.7.4 in my case)


acdha:
@m-thielen Can you add a test for this to prevent regressions in the future? Thanks!

m-thielen:
How do you setup testing, maybe just the whoosh backend?

acdha:
That's fine in this case — https://django-haystack.readthedocs.io/en/latest/running_tests.html describes the process of running tests, which in practice might be running `test_haystack/run_tests.py` from your working branch. By now the test suite should run only the tests for the available search engines so it shouldn't be necessary to follow the instructions to start Solr if all you're testing is Whoosh — not having the Solr or ElasticSearch clients installed in your python path should cause those test suites to be skipped.

Afnarel:
Hi! I stumbled upon this PR, and it seems this has been added since it was opened, right?
image

PR #1551, opened 2017-09-15
Uuid pk fix

### LINKED ISSUE #1543: Search (ala SearchView) doesn't work with UUID pk's(Non Int pk's in general)
* [x] Tested with the latest Haystack release

## Expected behaviour
You could use UUID pk's for models in django.  Setup a SearchView and it all works.

## Actual behaviour
You can see the number of results with sqs.all().count() in for example your SearchForm Class.  But the moment you try to instantiate and get all the results...it evaporates.  No results available when you try to iterate over the list.

1.Use UUID's as primary keys on django models that are indexed
2.Create a SearchView and SearchForm that use the Model.
3.Try to use the sqs in the html template through object_list OR a paginated result set (page_obj.object_list)

## Configuration

* Operating system version:Ubuntu 16.04
* Search engine version:6.5
* Python version:3.5.2
* Django version:1.11.4
* Haystack version:master branch and earlier

## Root Cause Analysis
Still in process but it looks like post_process_results of query.py:SearchQuerySet tries to force an int() conversion on the pk causing a problem...or it's involved somehow.  resulting in every result being discarded as an ignored result silently

Still working out the exact nature.

This took me forever to figure out.  I thought I had it working before...I don't know what changed to make it stop.

RabidCicada:
This is to fix #1543.  The primary purpose of the PR is to fix django-haystack to properly work with UUID primary key models for regular searching.

As fallout of the PR I have also implemented tests that actually use the UUIDMockModel (that was contributed previously but not actively used) and run it through a reasonable number of tests I think.

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -172,11 +172,14 @@ class SearchQuerySet(object):
 
         for result in results:
             if self._load_all:
-                # We have to deal with integer keys being cast from strings
+                
                 model_objects = loaded_objects.get(result.model, {})
+                # Try to coerce a primary key object that matches the models pk
+                # We have to deal with semi-arbitrary keys being cast from strings (UUID, int, etc)
                 if result.pk not in model_objects:
                     try:
-                        result.pk = int(result.pk)
+                        result_klass = type(next(iter(model_objects)))
+                        result.pk = result_klass(result.pk)
                     except ValueError:
                         pass
                 try:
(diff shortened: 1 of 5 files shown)


RabidCicada:
@acdha @bigjust Heyo.  Please could I get this merged in?  This completes full uuid support that was began in the last major PR #1504 (That added solr 6 support and began uuid support).

acdha:
Sorry - I've been out on parental leave and haven't been checking GitHub much

RabidCicada:
TY!  Hope all is well with the kiddos.

PR #1560, opened 2017-10-08
Update boost.rst

cvarelaruiz:
To fix the dict key that needs to be passed for Document boosting ('_boost' and not 'boost')

### DIFF
--- a/docs/boost.rst
+++ b/docs/boost.rst
@@ -57,7 +57,7 @@ See the :doc:`searchqueryset_api` docs for more details on using this method.
 Document Boost
 ==============
 
-Document boosting is done by adding a ``boost`` field to the prepared data
+Document boosting is done by adding a ``_boost`` field to the prepared data
 ``SearchIndex`` creates. The best way to do this is to override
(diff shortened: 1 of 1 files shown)


acdha:
Which search backend were you using? It looks like that key must have changed at some point and we’d need to either document that or check both

cvarelaruiz:
Hey, I'm using Whoosh for now and couldn't make it work with the dict key 'boost'. Changing it to '_boost' had the desired effect. 

Is that parameter engine dependent? I thought that was part of the layer of abstraction that Haystack provides. 

Cristian

bigjust:
dug into this a bit, and if i'm reading this correctly: 

https://github.com/django-haystack/django-haystack/blob/95719cf4f84707fa73d08596c52e229c3bbcf9e6/haystack/backends/whoosh_backend.py#L194-L196

might be the problem? either way, might need to be abstracted. I don't see anything supporting that document boosts aren't supported, unless that is simply no longer the case in 2.7 or whatever the current version is.

PR #1566, opened 2017-11-11
Clarifies need for HAYSTACK_DOCUMENT_FIELD

chrisbay:
This setting is needed when using a document field name other than `text`, which is not clear as-written.


denizdogan:
Would be of great benefit if this PR got reviewed and merged.

acdha (approved):

PR #1569, opened 2017-11-27
Checking for __iter__ not valid way to check is iterable

birdbrained:
Some objects have an `__iter__` method even though they are not iterable - `SimpleLazyObject` for ex.

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -153,7 +153,9 @@ class SearchField(object):
                 return current_objects.all()
             return []
 
-        elif not hasattr(current_objects, '__iter__'):
+        try:
+            current_objects = [obj for obj in current_objects]
+        except:
             current_objects = [current_objects]
 
         return current_objects


haystack/fields.py
```diff
@@ -153,7 +153,9 @@ def get_iterable_objects(cls, current_objects):
                 return current_objects.all()
             return []
 
-        elif not hasattr(current_objects, '__iter__'):
+        try:
+            current_objects = [obj for obj in current_objects]
```
klass-ivan: Is there any example of iterable without `__iter__` method? All basic collections, generators and iterators have it.
I only noticed that 'string' object has no `__iter__` in python2 (it is present in python3) but string is an atomic value in terms of fields. Also, checking for `__iter__` is quite common practice to check for iteration ability.
```
>>> list().__iter__
<method-wrapper '__iter__' of list object at 0x7fb27115f908>
>>> set().__iter__
<method-wrapper '__iter__' of set object at 0x7fb271185d00>
>>> dict().__iter__
<method-wrapper '__iter__' of dict object at 0x7fb27116bb40>
>>> tuple().__iter__
<method-wrapper '__iter__' of tuple object at 0x7fb27126d050>
>>> (xrange(10)).__iter__
<method-wrapper '__iter__' of xrange object at 0x7fb271166648>
>>> (xrange(10)).__iter__().__iter__()
<rangeiterator object at 0x7fb271169480>
>>> def gen():
...     for x in xrange(10): yield x
... 
>>> gen().__iter__()
<generator object gen at 0x7fb271155820>
>>> gen().__iter__().__iter__()
<generator object gen at 0x7fb271155a50>
```
If there is any common case of such a iterable without `__iter__`, I think that should be shown in test.

birdbrained: This issue was uncovered a lttle while ago - I can't remember the object that was being given here - if I recall it was a proxy object to a queryset or something to that effect - the proxy itself does not have an __iter__ method, but when passed to a comprehension it behaves as one would expect.

In addition, the above change makes the logic here more pythonic as its a common idiom in python to ask for forgiveness not permission. https://docs.python.org/3.6/glossary.html#term-eafp

klass-ivan:
Failing test:
`SearchFieldTestCase.test_get_iterable_objects_with_list_stays_the_same` in `test_haystack.test_fields`

birdbrained:
Oh I wrote which object in that original comment - looks like it was a `SimpleLazyObject`.

haystack/fields.py
```diff
@@ -153,7 +153,9 @@ def get_iterable_objects(cls, current_objects):
                 return current_objects.all()
             return []
 
-        elif not hasattr(current_objects, '__iter__'):
+        try:
+            _ = [obj for obj in current_objects]
```
klass-ivan: This will consume all the possible generators:

```
class Test(object):
    @property
    def test(self):
        for x in '12345': yield x

current_objects = Test().test
_ = [c for c in current_objects]
print(', '.join(current_objects))
```

acdha: Yes — I think the way to handle this would need to be something like this:

```python
try:
    iter(current_objects)
except TypeError:
    current_objects = [current_objects]
```

jnm: I hadn't realized this PR already existed, but I used the `iter(current_objects)` approach suggested by @acdha for PR #1663, back in 2019.

PR #1588, opened 2018-01-29
make BaseInput.__repr__ for in python3

bit:
remove call to __unicode__


acdha:
That's definitely some old code deserving a cleanup. Can we get rid of the `__str__`, too?

acdha:
The test failures on Python 3 should go away if you rebase against the current master branch

PR #1590, opened 2018-02-16
Fixed index re-obtaining for every result entry field when searching by ElasticSearch backend

klass-ivan:
Looks like one line of code in elasticsearch `_process_results` is misplaced.
It results in obtaining index on every! field per each result entry. That's useless overhead anyway. Also it may work much slower if you have overridden .get_index in your custom UnifiedIndex

### DIFF
--- a/haystack/backends/elasticsearch_backend.py
+++ b/haystack/backends/elasticsearch_backend.py
@@ -625,8 +625,8 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
             model = haystack_get_model(app_label, model_name)
 
             if model and model in indexed_models:
+                index = source and unified_index.get_index(model)
                 for key, value in source.items():
-                    index = unified_index.get_index(model)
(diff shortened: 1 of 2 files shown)


klass-ivan:
Just want to share some performance estimations:
There is an index containing 35 fields, total about 200 records in ES.
I've dual-core laptop CPU Intel Core i7 7500U.
Here is some timing of getting a `list(search_queryset)`:

> 
> Without patch, ES just started
> 0.14482
> 0.131979
> 0.153539
> 0.13158
> 0.181441
> 
> Applied patch, ES re-started
> 0.056528
> 0.068232
> 0.065078
> 0.085449
> 0.069956
> 
> Patch removed, ES not re-started
> 0.139259
> 0.136195
> 0.123292
> 0.149413
> 0.140626
> 

So as for me a better placement of this line is **about 2x performance speed-up**!
The more fields in the index, the more significant it will be.

PR #1596, opened 2018-03-14
Fixed collection of deep attributes through m2m relation

klass-ivan:
When model `A` has a Many-to-Many relation `B` which has another relation `C`, the `A` index field with `model_attr='b__c__some_attr'` works improperly - only first item is collected.

That's because in `resolve_attributes_lookup` there is a `return` for first object:

    for current_object in current_objects:
            # ...
            if len(attributes) > 1:
                return .... # resolved only for current_object

As result, further objects are not looked into.


klass-ivan:
@acdha  Thank you for merging this in. Would you mind building a minor release with this change? Thanks.

PR #1601, opened 2018-04-18
__len__() should return >= 0  error fixed

AnjaneyuluBatta505:
python2.7/site-packages/haystack/query.py” in count

522.         return len(self)
Exception Type: ValueError at /search/ Exception Value: __len__() should return >= 0 Request information: USER: AnonymousUser

### DIFF
--- a/haystack/query.py
+++ b/haystack/query.py
@@ -90,7 +90,7 @@ class SearchQuerySet(object):
                 self._result_count = 0
 
         # This needs to return the actual number of hits, not what's in the cache.
-        return self._result_count - self._ignored_result_count
+        return max([self._result_count - self._ignored_result_count, 0])
 
     def __iter__(self):
         if self._cache_is_full():


acdha:
Can you explain the circumstances where you get a negative number of results? Ideally this would have a test-case, too.

AnjaneyuluBatta505:
@acdha In some cases the some of the indexes may delete(or not properly indexed) or which may not have `pk`  which will result in  
increase of `_ignored_result_count` and if query match count returns `0` then it will result in negative value. According to python datamodel `__len__` should not return negative values.
So, It needs to be fixed.
Thanks!

frankystone:
@AnjaneyuluBatta505 

> In some cases the some of the indexes may delete(or not properly indexed) or which may not have pk

Wouldn't be the index then be in a malformed state? Searching may lead into wrong search results, or nothing will be found, although the search term should be found?

AnjaneyuluBatta505:
@frankystone 
searching will return correct results because if index's are malformed then they will not be cosidered. So, we can return 0 results. If we did not fix '__len__' then it will raise ValueError. So, it needs to be fixed.

AnjaneyuluBatta505:
@frankystone  and @acdha  please approve the merge request.

acdha:
The PR still could use some tests

sebhaase:
#1601 ist not working for me:  no error anymore, but also no results found

AnjaneyuluBatta505:
@sebhaase try the command below
```shell
python manage.py update_index  --remove
```
It will update the new records and removes the deleted records.

sebhaase:
@AnjaneyuluBatta505 you misunderstood. Nothing was deleted. I'm saying #1601 causes that I´m not seeing the correct answer.  Old haystack version still works for me - `update_index` makes no difference.

AnjaneyuluBatta505:
@sebhaase  I've tried it. It's working for me.

frankystone:
The remove option seems to work also for me.

But it is just like rebuild_index, at least it is just that slow :(

gavinwahl:
This is still happening

useHTML5:
actual

Issue #1571, opened 2017-12-01
DistutilsError: Could not find suitable distribution for Requirement.parse('setuptools-scm')


palewire:
I'm getting the following `setuptools_scm` error when trying to install via the latest version of pip in a virtual environment. As you can see in the code below, setuptools_scm does appear to be installed. I'm trying to figure out why it's not linking up with your library. If you have any ideas, please let me know.

```bash
(himmel.latimes.com)datadesk@himmel:/apps/himmel.latimes.com$ pip freeze | grep setuptools-scm
setuptools-scm==1.15.6
(himmel.latimes.com)datadesk@himmel:/apps/himmel.latimes.com$ python
Python 2.7.3 (default, Aug  1 2012, 05:14:39) 
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools_scm
>>> exit()
(himmel.latimes.com)datadesk@himmel:/apps/himmel.latimes.com$ pip install django-haystack
Collecting django-haystack
  Using cached django-haystack-2.6.1.tar.gz
    Complete output from command python setup.py egg_info:
    Couldn't find index page for 'setuptools_scm' (maybe misspelled?)
    No local packages or download links found for setuptools-scm
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-do137B/django-haystack/setup.py", line 68, in <module>
        setup_requires=['setuptools_scm'],
      File "/usr/lib/python2.7/distutils/core.py", line 112, in setup
        _setup_distribution = dist = klass(attrs)
      File "/apps/himmel.latimes.com/local/lib/python2.7/site-packages/distribute-0.6.24-py2.7.egg/setuptools/dist.py", line 221, in __init__
        self.fetch_build_eggs(attrs.pop('setup_requires'))
      File "/apps/himmel.latimes.com/local/lib/python2.7/site-packages/distribute-0.6.24-py2.7.egg/setuptools/dist.py", line 245, in fetch_build_eggs
        parse_requirements(requires), installer=self.fetch_build_egg
      File "/apps/himmel.latimes.com/local/lib/python2.7/site-packages/distribute-0.6.24-py2.7.egg/pkg_resources.py", line 576, in resolve
        dist = best[req.key] = env.best_match(req, self, installer)
[...]

PR #1603, opened 2018-04-23
mv setup_requires to install requires

Friday21:
according to this [issue](https://github.com/pypa/pip/issues/4880) and this [one](https://github.com/django-haystack/django-haystack/issues/1571),  when install pypi with my custom pypi, it will raise DistutilsError, because my server can't access https://pypi.org/simple/,  because setuptools uses it's own logic to find the distributions mentioned in setup_requires. And, that logic is unable to fetch the distribution. so I move it to  install_requires.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ except ImportError:
 
 install_requires = [
     'Django>=1.11',
+    'setuptools_scm',
 ]
 
 tests_require = [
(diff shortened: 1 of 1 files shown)


aerostitch:
It would be great if we could have this one merge, I believe it would fix 2 RC bugs we have in Debian: https://bugs.debian.org/896296 and https://bugs.debian.org/896263
Thanks! :)

claudep:
`setup.py` was changed multiple times since you suggested this PR. If you still have an issue, feel free to reopen a new PR.

PR #1646, opened 2018-10-05
Add support for custom field types w/ Solr backend

tc77:
The Solr backend supported only `text_en` for `CharField` which is not appropriate for some use-cases. There you can now derive a custom Field with a get_field_type method which delivery the required field-type.

### DIFF
--- a/haystack/backends/solr_backend.py
+++ b/haystack/backends/solr_backend.py
@@ -544,23 +544,23 @@ class SolrSearchBackend(BaseSearchBackend):
             if field_class.document is True:
                 content_field_name = field_class.index_fieldname
 
-            # DRL_FIXME: Perhaps move to something where, if none of these
-            #            checks succeed, call a custom method on the form that
-            #            returns, per-backend, the right type of storage?
-            if field_class.field_type in ['date', 'datetime']:
-                field_data['type'] = 'date'
-            elif field_class.field_type == 'integer':
-                field_data['type'] = 'long'
-            elif field_class.field_type == 'float':
-                field_data['type'] = 'float'
-            elif field_class.field_type == 'boolean':
-                field_data['type'] = 'boolean'
-            elif field_class.field_type == 'ngram':
-                field_data['type'] = 'ngram'
-            elif field_class.field_type == 'edge_ngram':
-                field_data['type'] = 'edge_ngram'
-            elif field_class.field_type == 'location':
-                field_data['type'] = 'location'
+            try:
+                field_data['type'] = field_class.get_field_type()
+            except AttributeError:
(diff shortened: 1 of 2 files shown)


haystack/backends/solr_backend.py
```diff
@@ -544,23 +544,23 @@ def build_schema(self, fields):
...
-                field_data['type'] = 'boolean'
-            elif field_class.field_type == 'ngram':
-                field_data['type'] = 'ngram'
-            elif field_class.field_type == 'edge_ngram':
-                field_data['type'] = 'edge_ngram'
-            elif field_class.field_type == 'location':
-                field_data['type'] = 'location'
+            try:
+                field_data['type'] = field_class.get_field_type()
+            except AttributeError:
```
acdha: I like this approach a lot more than what we've been doing but was wondering whether we might want to do something like a `hasattr` check to avoid the possibility of a silent failure if someone's `get_field_type` hit an `AttributeError`.

haystack/backends/solr_backend.py
```diff
@@ -663,23 +663,24 @@ def build_schema(self, fields):
...
-            elif field_class.field_type == "boolean":
-                field_data["type"] = "boolean"
-            elif field_class.field_type == "ngram":
-                field_data["type"] = "ngram"
-            elif field_class.field_type == "edge_ngram":
-                field_data["type"] = "edge_ngram"
-            elif field_class.field_type == "location":
-                field_data["type"] = "location"
+            if (hasattr(field_class, "get_field_type") and 
+                    callable(getattr(field_class, "get_field_type"))):
```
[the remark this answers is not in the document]
acdha: Right, which is why I was wondering whether this might be better as a simple

```python
if hasattr(field_class, "get_field_type"):
```

since the extra check isn't really adding anything and if someone does make an uncallable property named `get_field_type` this won't delay them realizing their mistake.

haystack/backends/solr_backend.py
```diff
@@ -663,23 +663,23 @@ def build_schema(self, fields):
...
-            elif field_class.field_type == "boolean":
-                field_data["type"] = "boolean"
-            elif field_class.field_type == "ngram":
-                field_data["type"] = "ngram"
-            elif field_class.field_type == "edge_ngram":
-                field_data["type"] = "edge_ngram"
-            elif field_class.field_type == "location":
-                field_data["type"] = "location"
+            if hasattr(field_class, "get_field_type"):
+                field_data["type"] = getattr(field_class, "get_field_type")()
```
acdha:
```suggestion
                field_data["type"] = field_class.get_field_type()
```

PR #1656, opened 2019-01-07
Move import into function to avoid GDAL as a global dependency

### LINKED ISSUE #1631: App server won't start if GDAL library is not setup
* [x] Tested with the latest Haystack release

## Expected behaviour
Setting up GDAL library should be optional, if someone doesn't want to use LocationField

## Actual behaviour

django.contrib.gis.geos.Point import is global in fields.py which depends on GDAL library and it throws an exception: ImproperlyConfigured: Could not find the GDAL library

In the version Haystack 2.8.1 this is a local import in the convert function of the LocationField, which solves the issue.

1. Make sure that GDAL library is not setup for your app, runserver will fail with the above mentioned exception.

## Configuration

* Operating system version:  macOS 10.13.4
* Search engine version: Elasticsearch 5.5.3
* Python version: 2.7.13
* Django version: 1.11.13
* Haystack version: Latest Master from Commit - bbcd038a96a4572bafa7f47bdf9eb0d345a2445b

mpauly:
Hi, 
this should fix #1631 - The problem arises because in commit 39c2adcd720485410c289601d696473027539358 an import was moved from within a function to the top level in `haystack/fields.py`. The import references `django.contrib.gis.geos`. That import in turn causes a lot of other imports that end in a call to GDAL. This call fails if GDAL is not installed. The fix is simple: Just move the import back to the function. The fix is tested in one of our installations (without GDAL).
Cheers,
Martin

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import re
 from inspect import ismethod
 
-from django.contrib.gis.geos import Point
 from django.template import loader
 from django.utils import datetime_safe, six
 
@@ -260,6 +259,7 @@ class LocationField(SearchField):
         return "%s,%s" % (pnt_lat, pnt_lng)
 
     def convert(self, value):
+        from django.contrib.gis.geos import Point
         from haystack.utils.geo import ensure_point
 
         if value is None:
--- a/test_haystack/elasticsearch2_tests/test_backend.py
+++ b/test_haystack/elasticsearch2_tests/test_backend.py
@@ -10,7 +10,6 @@ from decimal import Decimal
 import elasticsearch
 from django.apps import apps
 from django.conf import settings
-from django.contrib.gis.geos import Point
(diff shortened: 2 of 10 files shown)


acdha:
Thanks — I had started this in https://github.com/acdha/django-haystack/tree/defer-gdal-import which has a few other places where the GIS imports needed to be deferred.

mpauly:
Ah, great. I wasn't sure if I should also modify the tests and decided against it. Do you have any clue when you will find time to merge your branch?

acdha:
My branch isn't finished yet — it's on the TODO list but I haven't had much time for OSS. If you want to help get it to the point where the tests are passing I'd be happy to merge it.

mpauly:
I now moved all the imports from `django.contrib.gis.geos` to the functions in which they are used. The tests seem to be passing. Additionally I updated the two regex, that you updated as well as.

acdha:
Thanks for picking this up — I'll look into a point release soon since #1631 had a few people blocked on it

PR #1676, opened 2019-05-25
Replace type:'string' with type'text' for elasticsearch 5 compaitbility.

msapiro:
The `string` type is removed in elasticsearch 5. There is an automatic conversion to `text` but this doesn't support parameters like `boost`.

Attempts to PUT `_mapping/modelresult` with items like `'subject': {'type': 'string', 'analyzer': 'snowball', 'boost': 1.25},` result in a 400 error with the exception `{'error': {'root_cause': [{'type': 'illegal_argument_exception', 'reason': 'The [string] type is removed in 5.0 and automatic upgrade failed because parameters [boost] are not supported for automatic upgrades. You should now use either a [text] or [keyword] field instead for field [subject]'}], 'type': 'illegal_argument_exception', 'reason': 'The [string] type is removed in 5.0 and automatic upgrade failed because parameters [boost] are not supported for automatic upgrades. You should now use either a [text] or [keyword] field instead for field [subject]'}, 'status': 400}`

This PR modifies `backends/elasticsearch5_backend.py` to change `string` to `text` in the imported `backends/elasticsearch_backend.py` settings `DEFAULT_FIELD_MAPPING` and `FIELD_MAPPINGS`

### DIFF
--- a/haystack/backends/elasticsearch5_backend.py
+++ b/haystack/backends/elasticsearch5_backend.py
@@ -28,6 +28,15 @@ except ImportError:
                             installation of 'elasticsearch>=5.0.0,<6.0.0'. \
                             Please refer to the documentation."
     )
+# We need to override some elasticsearch_backend settings as string type is
+# gone in elasticsearch 5.
+from haystack.backends import elasticsearch_backend
+elasticsearch_backend.DEFAULT_FIELD_MAPPING = {
+    "type": "text", "analyzer": "snowball"}
+elasticsearch_backend.FIELD_MAPPINGS["edge_ngram"] = {
+    "type": "text", "analyzer": "edgengram_analyzer"}
+elasticsearch_backend.FIELD_MAPPINGS["ngram"] = {
+    "type": "text", "analyzer": "ngram_analyzer"}
 
 
 class Elasticsearch5SearchBackend(ElasticsearchSearchBackend):


msapiro:
There are a number of failed tests using elasticsearch 5. I suspect even without the changes in this PR, several elasticsearch 5 tests will fail although they may be different failures.

Some of the failures may because things that used to return `type: "string"` now return `type: "text"`, but others may be because a type should be `keyword` rather than `text`. Clearly a proper fix is more complex than this PR, and I don't yet know how to fix it. The changes in the PR work for me, but my model only has `indexes.CharField`, `indexes.DateTimeField` and `indexes.MultiValueField` so it is fairly simple.

cclauss:
Are these changes still needed?  If so, please rebase on the current `master`.

msapiro:
> Are these changes still needed? If so, please rebase on the current `master`.

They are still needed. I note this issue was fixed in django-haystack/haystack/backends/elasticsearch7_backend.py by redefining DEFAULT_FIELD_MAPPING and FIELD_MAPPINGS. I think this is a better approach so I have changed this PR to apply that fix to django-haystack/haystack/backends/elasticsearch5_backend.py as well.

jochengcd:
for consistency with the change #1965 for ES7 these c/should be class variables and used in build_schema with self.

PR #1682, opened 2019-07-29
Queryset cache using sparse list update

wetneb:
This is an update of #1451, using the released version of `sparse_list`.

I will test this in production and report back.


wetneb:
Tests are currently failing because of https://github.com/johnsyweb/python_sparse_list/issues/12.

PR #1689, opened 2019-10-15
using six replace django six

the5fire:
ref: https://docs.djangoproject.com/en/dev/releases/3.0/#removed-private-python-2-compatibility-apis


acdha:
It might be better to remove six at this point along with Python 2 support

the5fire:
@acdha Agree.

But that will be a big change. We need to modify the version that depends on Django to be >= 2.0.

PR #1698, opened 2019-12-03
Travis CI: upgrade to running on Xenial

acdha:
Long-overdue upgrade to the Travis CI runner

PR #1714, opened 2020-01-24
Travis: upgrade to bionic

sk1p:
Supersedes https://github.com/django-haystack/django-haystack/pull/1698


sk1p:
It seems that 1) elasticsearh 1.7 is not installable on bionic, 2) elasticsearch 2.x tests fail to connect to elasticsearch (see [this job](https://travis-ci.org/django-haystack/django-haystack/jobs/641472183) for an example, search for elasticsearch2_tests), and 3) elasticsearch 5.x [works fine](https://travis-ci.org/django-haystack/django-haystack/jobs/641472234?utm_medium=notification).

Is there any schedule for removing support for older elasticsearch? Otherwise, it may be possible to
[run old elasticsearch versions via docker](https://hub.docker.com/_/elasticsearch?tab=tags&page=1&ordering=-last_updated&name=1.7.6)

acdha:
I would be fine removing ancient versions of ElasticSearch but otherwise I would be a strong +1 on using Docker for that since it's not a core problem for us. Much of Haystack's test setup predates Docker being an option and there's plenty of room for improvement — I was thinking about doing the same for Solr, too.

sk1p:
@acdha running via docker seems to work, even for the ancient versions. I think removing elasticsearch 1.x support should be done in a separate PR, as it also requires some more cleanup in the backend itself.

For solr: I can give it a try, should the `start-solr-test-server.sh` script be replaced with a small script that just calls docker?

sk1p:
Solr via docker should now at least run - waiting for the current slow run to finish  - there are a few tests that fail, tough. One is `test_build_schema` which expects to find the `mgmnt/conf` directory, which it can't find, because it is created inside the container:

```
FAIL: test_build_schema (test_haystack.solr_tests.test_solr_management_commands.ManagementCommandTestCase)
----------------------------------------------------------------------
  File "/home/alex/source/django-haystack/test_haystack/solr_tests/test_solr_management_commands.py", line 252, in test_build_schema
    os.path.isdir(conf_dir), msg="Expected %s to be a directory" % conf_dir
AssertionError: False is not true : Expected /home/alex/source/django-haystack/test_haystack/solr_tests/server/solr/server/solr/mgmnt/conf to be a directory

----------------------------------------------------------------------
```


The other kind of failure looks like this:

```
ERROR: test_build_schema_wrong_backend (test_haystack.solr_tests.test_solr_management_commands.ManagementCommandTestCase)
----------------------------------------------------------------------
  File "/home/alex/source/django-haystack/test_haystack/solr_tests/test_solr_management_commands.py", line 65, in setUp
    self.ui.build(indexes=[self.smmi])
  File "/home/alex/source/django-haystack/haystack/utils/loading.py", line 260, in build
    self.collect_fields(index)
  File "/home/alex/source/django-haystack/haystack/utils/loading.py", line 270, in collect_fields
    % (self.document_field, index)
haystack.exceptions.SearchFieldError: All 'SearchIndex' classes must use the same 'Th3S3cr3tK3y' fieldname for the 'document=True' field. Offending index is '<test_haystack.solr_tests.test_solr_management_commands.SolrMockSearchIndex object at 0x7f77349dca60>'.
```

Not sure what is up with that one. Maybe some state from `test_build_schema` that is not cleaned up? It's been a while since I used solr, so I would appreciate help with these.

If the rest of the PR looks good to you, I can also split the solr changes into a new PR.

sk1p:
> Maybe some state from test_build_schema that is not cleaned up?

That was it, wrapping in `try..finally` fixed these.

acdha:
I think leaving it in this PR is reasonable since you're working on this now.

re: Solr schema, I'm not sure you have a preference for using a volume for that (or `docker cp`) or just to use the Solr Schema API.

sk1p:
> I think leaving it in this PR is reasonable since you're working on this now

:+1: 

> re: Solr schema, I'm not sure you have a preference for using a volume for that (or `docker cp`) or just to use the Solr Schema API.

Hmm, using a volume leads to permission problems, because solr is using as a different UID in the container than the user running the tests. I can get the container to write to the volume by `chmod 777`ing it, but then the test code can't write there...

I'm not sure what approach to take with either `docker cp` or using the schema API - I can retrieve the schema without problem, but to circumvent the permission problem above, the management command would have to be changed to use the schema API to replace the schema (and config?), and I don't think that is actually possible.

sk1p:
pypy build failure:

```
OSError: Cannot load library libgdal.so.20: /usr/lib/libgdal.so.20: undefined symbol: sqlite3_column_table_name
```

Since libgdal20 installs libspatialite7 and thus libsqlite3-0, the symbols _should_ be available, but can't be found here. Maybe because the pypy3 version by travis is built against a different sqlite3? One can only guess...

sk1p:
I've reverted the solr+docker related changes as I'm not sure I'll have time to finish them soon. If someone else wants to pick them up, they are cherry-picked onto https://github.com/sk1p/django-haystack/tree/solr-docker

PR #1720, opened 2020-02-12
Drop support for Python 2 and old versions of Django. Add support for Django 3

solarissmoke:
Hi there!

This PR does the following:

1. Drops support for Python 2 which is now end of life. This includes dropping all `__future__` imports and reliance on `six`, and replacing all Python 2 specific code (e.g., `__unicode__`) with the Python 3 equivalents (`__str__`).

2. Drops support for all versions of Django prior to version 2.2, [as recommended by the Django project](https://docs.djangoproject.com/en/3.0/releases/3.0/#third-party-library-support-for-older-version-of-django).

3. Adds support for Django 3.0.

4. Adds support for Python 3.8.

5. Substantially DRY out the `tox` configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
     - 3.5
     - 3.6
     - 3.7
+    - 3.8
     - pypy3
 services:
     - docker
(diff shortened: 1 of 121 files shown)


solarissmoke:
Thanks @acdha ! If you are able to cut a release on pypi in the next couple of weeks that would be appreciated.

acdha:
I dropped this just now: https://pypi.org/project/django-haystack/3.0b1/

stefanocoding:
Thanks for the added support.

In this line https://github.com/django-haystack/django-haystack/blob/master/haystack/backends/solr_backend.py#L525, `raw_result[DJANGO_CT]` returns a list instead of a string, which raises an error because of the call to `split()`. 

My setup: Django 3.1.1, Solr 8.6.2 and Haystack 3.0b1.

PR #1737, opened 2020-06-17
Whoosh official source has moved

homerjonathan:
Official Whoosh source has moved to https://github.com/whoosh-community/whoosh

No change to source just documentation so no testing required.


acdha:
Thanks!

PR #1743, opened 2020-10-01
Fix Whoosh bitbucket link (404) with Github link

cb109:
## Fixing a broken link in the README.md

The original https://bitbucket.org/mchaput/whoosh/ is not available anymore, but the same user seems to have migrated it to Github, so just linking to that instead: https://github.com/mchaput/whoosh/


dulmandakh (approved):
@acdha LGTM

acdha (approved):

PR #1745, opened 2020-10-08
Remove remaining Python2 logic

Cabalist:
Hey there,

I've just cleaned out some old Python2 compatibility stuff and modernized the super calls for Python3.   Nothing major.   

I forked the code to fix some warnings in Django3 but I see they were already cleaned up in master.  :) Thank you for staying on top of those.


Cabalist:
It is worth pointing out that in the code below the super call references the inherited class instead of the class itself and so has slightly different behaviour than usual.   If that is expected that is fine but it *may* be a bug.

https://github.com/django-haystack/django-haystack/blob/1cc82885e287fe33bb937047137da3d803072684/haystack/query.py#L707-L718

Cabalist:
The code duplication is not my doing obviously.  :)

dulmandakh:
@Cabalist could you please resolve the conflict

Cabalist:
Done.  :)

Cabalist:
This is ready to merge.   :)

dulmandakh (approved):
@acdha LGTM, please review and merge

acdha (approved):

PR #1751, opened 2020-12-23
minutes granularity for update index command

yeago:


acdha:
Closing as a duplicate of #1752. In the future, it's probably easier just to push the extra commits to the same branch to avoid duplication.

haystack/management/commands/update_index.py
```diff
@@ -248,6 +255,12 @@ def handle(self, **options):
         elif self.verbosity > 1:
             LOG.setLevel(logging.INFO)
 
+        if (minutes and age) or (minutes and start_date) or (age and start_date):
+            raise NotImplementedError("Minutes / age / start date options are mutually exclusive")
```
acdha, line 259: This should call [`parser.error`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) instead for consistency.

```suggestion
            parser.error("Minutes / age / start date options are mutually exclusive")
```

It could be further simplified by using [`ArgumentParser.add_mutually_exclusive_group()`](https://docs.python.org/3/library/argparse.html#mutual-exclusion) so the help will automatically show these as exclusive options.

haystack/management/commands/update_index.py
```diff
@@ -248,6 +255,12 @@ def handle(self, **options):
...
+
+        if minutes is not None:
+            self.start_date = now() - timedelta(minutes=int(minutes))
```
acdha, line 262: Since the argument is defined as an integer, the extra `int()` call probably isn't necessary.

yeago: ah yeah was just consistency with age but i suppose we could dump both

acdha (approved):

PR #1754, opened 2021-01-20
 Improve the performance of ORM calls

django-doctor:
using `queryset.count()` is more efficient than checking `len(queryset)` [read more](https://django.doctor/advice/C9000)

I found this problem during auditing your codebase. [Read the full report](https://django.doctor/django-haystack/django-haystack)

### DIFF
--- a/test_haystack/test_indexes.py
+++ b/test_haystack/test_indexes.py
@@ -245,15 +245,15 @@ class SearchIndexTestCase(TestCase):
         self.assertTrue(isinstance(self.cmi.fields["extra"], indexes.CharField))
 
     def test_index_queryset(self):
-        self.assertEqual(len(self.cmi.index_queryset()), 3)
+        self.assertEqual(self.cmi.index_queryset().count(), 3)
 
     def test_read_queryset(self):
-        self.assertEqual(len(self.cmi.read_queryset()), 2)
+        self.assertEqual(self.cmi.read_queryset().count(), 2)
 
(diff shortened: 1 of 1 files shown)


yeago:
i am confused the guidance says don't use count() but the diff changes things to count() ...?

django-doctor:
@yeago you're right to be confused as I screwed up the PR description. Will fix it shortly

acdha:
@django-doctor The commit still shows it changing fast code to slow code as covered by the description.

yeago:
i think its hypothetically faster but given that its test data i don't think it addresses efficiencies that will be seen in production... more just better practices for the test codebase

richtier:
hi @acdha, author of @django-doctor here. Just popping by to clear up faster vs slower. From the [Django docs](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#when-querysets-are-evaluated).

> Note: If you only need to determine the number of records in the set (and don’t need the actual objects), it’s much more efficient to handle a count at the database level using SQL’s SELECT COUNT(*). Django provides a count() method for precisely this reason.

happy to reopen? If not I'll take my leave :)

acdha:
@richtier Yes, I'm well aware of the docs and performance characteristics. I closed this before the GitHub description had updated to the current version. It probably doesn't make a measurable difference with the tests using a local SQLite with a tiny number of records but there's no reason not to merge it.

PR #1759, opened 2021-02-27
Allow sorting for admin views

mpauly:
Hey,
the admin changelist in `SearchModelAdminMixin` previously did not allow to sort. With this change in `SearchQuerySet` this becomes possible.
Adding this without further changes would alter the behaviour for some users as the sqs is now sorted. 
To avoid this I modified `SearchModelAdminIndex.get_ordering` to not sort if a query is given. This is a bit of a hack, but should keep full backwards compatibility. Alternatively one could just drop this change.
Tested with Xapian.
Cheers, Martin

### DIFF
--- a/haystack/admin.py
+++ b/haystack/admin.py
@@ -30,6 +30,10 @@ class SearchChangeList(ChangeList):
             .load_all()
         )
 
+        ordering = self.model_admin.get_ordering(request)
+        if ordering:
+            qs = qs.order_by(*ordering)
+
         paginator = Paginator(sqs, self.list_per_page)
         # Get the number of objects, with admin filters applied.
         result_count = paginator.count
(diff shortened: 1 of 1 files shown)


mpauly:
It took me a while to get django-haystacks test suite to run, but now tests should be passing. The failure above seems to originate in Travis CI problems...

acdha (commented):
I'm definitely mixed on having that `-pk` check because that would prevent intentional use. I'm wondering whether it's worth using some kind of default sort constant or simply looking saying that it's not the end of the world if the ordering changes since there were no guarantees previously.

mpauly:
I see the point about intentional use. On the other hand setting `-pk` as a default order field might be problematic for people who do not have a field `pk` in their index... 
The problem is that given the current implementation in Django there is no way to tell if `-pk` is returned intentionally or as a default. So maybe it is best to just overwrite the `get_ordering` functionality of `ChangeList`, and implement custom logic that just works with a field `ordering` on the `SearchModelAdminMixin`. What do you think about that?

PR #1769, opened 2021-04-11
add GitHub Actions as CI

dulmandakh:
This PR adds GitHub Actions as CI, which will reduce CI time and improve developer experience. Added Django 3.2 and Python 3.9, also removed Django 3.0 which is EOL.

@acdha please review and merge


acdha (approved):
This looks great! Tempted to remove Travis at the same time.

PR #1780, opened 2021-04-14
flake8

dulmandakh:
Introduce flake8 check, and fix issues.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,17 @@
+name: flake8
+
+on: [pull_request, push]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Install tools
+      run: pip install flake8
+    - name: Run flake8
+      run: flake8 example_project haystack
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -12,12 +12,6 @@ jobs:
       with:
         python-version: 3.9
     - name: Install tools
-      run: pip install flake8 \
-               flake8-assertive \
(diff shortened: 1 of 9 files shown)


acdha:
Might be worth adding a few other flake8 plugins — I had enabled these on pysolr:

https://github.com/django-haystack/pysolr/blob/master/.pre-commit-config.yaml#L36-L39

These days I'd probably use this list:

```
                - flake8-assertive
                - flake8-builtins
                - flake8-comprehensions
                - flake8-logging-format
                - flake8-eradicate
                - flake8-bugbear
```

Might also be good to add isort and black while we're at it since we [tell people we're using those](https://github.com/django-haystack/django-haystack/blob/master/docs/contributing.rst#guidelines-for-contributing-code) and IIRC there's a GitHub action which can automatically apply the changes now.

dulmandakh:
@acdha I'm preparing another PR for black+isort because I didn't expected that you would ask them

dulmandakh:
@acdha also I didn't have time to fix all the issues from various plugins, so I added just flake8 to see if you would accept the change.

dulmandakh:
also run flake8 on tests, and there are so many issues to fix

acdha:
> also run flake8 on tests, and there are so many issues to fix

I think it'd make sense to ignore tests initially.

acdha:
Looking at this, I'm thinking we defer the logging format updates since those are ~50 separate errors.

dulmandakh:
@acdha let's merge this as is. I'll try to add flake8 plugins and fix issues.

acdha:
> @acdha let's merge this as is. I'll try to add flake8 plugins and fix issues.

Thanks - my first thought is that we should get Black + isort setup and then I'd probably want to review the current crop of flake8 warnings. There are a few general cleanups to be made but I've certainly seen worse. I only see one case of old Python2-isms (haystack/panels.py:43:59 has a dict iter* call) but otherwise it looks pretty straight forward.

PR #1786, opened 2021-04-20
add more flake8 plugins

dulmandakh:
add flake8 plugins mentioned in https://github.com/django-haystack/django-haystack/pull/1780

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ exclude=docs
 [flake8]
 line_length=88
 exclude=docs,tests
-ignore=E203, E501, W503, D
+ignore=E203, E501, E800, W503, D
 
 [options]
 setup_requires =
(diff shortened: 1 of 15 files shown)


setup.cfg
```diff
@@ -5,7 +5,7 @@ exclude=docs
 [flake8]
 line_length=88
 exclude=docs,tests
-ignore=E203, E501, W503, D
+ignore=E203, E501, E800, W503, D
```
acdha, line 8: `E800` is most of the point to flake8-eradicate. It looks like we have a mix of truly commented out code and some things like test suite code with comments showing values which happen to be valid Python. Think we should just defer eradicate for now?

acdha (approved):

setup.cfg
dulmandakh: removed flake8-eradicate

acdha (approved):

PR #1584, opened 2018-01-15
feat: expose a new option `ANALYZER`

zry656565:
- expose a new option `ANALYZER`, which enable to replace default analyzer
- so we can replace default StemmingAnalyzer with a configured analyzer, like `jieba.analyze.ChineseAnalyzer`.

```python
HAYSTACK_CONNECTIONS = {
    'default': {
        'ENGINE': 'haystack.backends.whoosh_backend.WhooshEngine',
        'PATH': os.path.join(PROJECT_DIR, 'whoosh_index'),
    },
    'default': {
        'ENGINE': 'haystack.backends.whoosh_backend.WhooshEngine',
        'PATH': os.path.join(PROJECT_DIR, 'whoosh_cn_index'),
        'ANALYZER': 'jieba.analyze.ChineseAnalyzer',
    },
}
```

Issue #327, opened 2011-02-25
Stemming (at least with Whoosh) should be disabled by default


rassie:
Currently, Haystack's Whoosh backend automatically uses the StemmingAnalyzer, which leads to funny results, especially with people's names and in non-english environments. One example is the search term "Sodenthaler" (a water brand in Germany) which is reduced to Term('text', u'sodenthal') thinking it might be an adjective in comparative form. Therefore, the result of searching for "Sodenthaler" is under certain circumstances empty, while a search for "Sodenthal" yields something useful.

Long story short: the following lines in backends/whoosh_backend.py would fix that for now:

...
            else:
                analyzer = None
                if getattr(settings, 'HAYSTACK_WHOOSH_STEMMING', False):
                    analyzer = StemmingAnalyzer()
                schema_fields[field_class.index_fieldname] = TEXT(stored=True, analyzer=analyzer)
....

The else line is line number 146 in my current installation (should be 1.1.0)

corro:
I am aware that this is an old issue, but I wanted you to know that this default behavior can indeed lead to strange search results. I did not expect this to be the default behavior and I even had to monkey patch the haystack whoosh backend in order to change this.

As a user, I would expect stemming to be disabled by default (like in whoosh itself) and to be configurable in the django settings (like @rassie proposed).

denizdogan:
Note that if you pass `analyzer=None` to a field, Whoosh will default to using `StandardAnalyzer`, so the solution isn't as simple as described in the first post. (Maybe this has changed in the ten years that have passed since creating the issue, I just know what the current state of Whoosh is.) Additionally, if `analyzer` is None and `lang` is NOT None, it will default to using `LanguageAnalyzer(lang)`. Again, this is in Whoosh, not in Haystack.

I've created a PR for a related change, #1789, and if that one is approved and merged, my suggestion for this issue is:

Let the default be `TEXT(analyzer=None)` and let the users figure out what that means from reading the Whoosh documentation. It's the only thing that makes sense to me. Haystack adding its own defaults to Whoosh's defaults is confusing to everyone and causes more issues. This change is not backward compatible, so it would be best suited for inclusion in a major release.

Thoughts on the above?

PR #1789, opened 2021-05-20
Allow per-field configuration of Whoosh analyzers

denizdogan:
Add a keyword argument, `analyzer`, for Whoosh fields to specify the analyzer for that field.

The default behavior remains the same, defaulting to `StemmingAnalyzer()` as the only analyzer. That's a default that I'm very keen on changing, but it's out of scope for this PR. (Issue #327)

I had a hard time figuring out how the unit tests were related, but I think I managed to add one in the right place. I decided to add an analyzed field for the authors in the bulk data file, which uses a regular expression to remove the digits from the names. In other words, daniel1, daniel2, and daniel3 are indexed as just "daniel".

Replaces PR's #452, #1037, and #1584. @acdha, if you find the time to look at this, please let me know what your thoughts are. This issue is pretty much a blocker for the project that I'm currently working on. Many thanks!

### DIFF
--- a/test_haystack/whoosh_tests/test_whoosh_backend.py
+++ b/test_haystack/whoosh_tests/test_whoosh_backend.py
@@ -29,7 +29,7 @@ class WhooshMockSearchIndex(indexes.SearchIndex, indexes.Indexable):
     pub_date = indexes.DateTimeField(model_attr="pub_date")
     name_analyzed = indexes.CharField(
         model_attr="author",
-        analyzer=SpaceSeparatedTokenizer() | SubstitutionFilter(r"\d+", "")
+        analyzer=SpaceSeparatedTokenizer() | SubstitutionFilter(r"\d+", ""),
     )
 
     def get_model(self):
--- a/test_haystack/whoosh_tests/test_whoosh_backend.py
+++ b/test_haystack/whoosh_tests/test_whoosh_backend.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.test import TestCase
 from django.test.utils import override_settings
 from django.utils.datetime_safe import date, datetime
+from whoosh.analysis import SpaceSeparatedTokenizer, SubstitutionFilter
 from whoosh.fields import BOOLEAN, DATETIME, KEYWORD, NUMERIC, TEXT
 from whoosh.qparser import QueryParser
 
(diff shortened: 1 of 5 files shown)


test_haystack/whoosh_tests/test_whoosh_backend.py
```diff
@@ -17,15 +18,19 @@
 from haystack.query import SQ, SearchQuerySet
 from haystack.utils.loading import UnifiedIndex
 
-from ..core.models import AFourthMockModel, AnotherMockModel, MockModel
-from ..mocks import MockSearchResult
-from .testcases import WhooshTestCase
+from test_haystack.core.models import AFourthMockModel, AnotherMockModel, MockModel
```
acdha, line 21: This change shouldn't have been necessary to get it to run — did you change it for aesthetic reasons or was it actually breaking local tests?

test_haystack/whoosh_tests/test_whoosh_backend.py
```diff
@@ -752,6 +757,11 @@ def test_scoring(self):
             ["0.40", "0.40", "0.40"],
         )
 
+    def test_analyzed_fields(self):  # TODO: rename to test_analyzed_fields
```
acdha, line 760: It looks like you already got to this TODO:

```suggestion
    def test_analyzed_fields(self):
```

test_haystack/whoosh_tests/test_whoosh_backend.py
```diff
@@ -752,6 +757,11 @@ def test_scoring(self):
...
+        self.sb.update(self.wmmi, self.sample_objs)
+        results = self.whoosh_search("name_analyzed:daniel")
```
acdha, line 762: This test doesn't appear to exercise the analyzer – at first glance it looks like we should be able to test it by adding some digits which should be stripped by the substitution filter:

```suggestion
        results = self.whoosh_search("name_analyzed:12345daniel67890")
```

acdha (approved):

test_haystack/whoosh_tests/test_whoosh_backend.py
denizdogan: My bad, I think that change slipped in because I was trying to help PyCharm resolve some warnings or errors. I'll revert this change!

test_haystack/whoosh_tests/test_whoosh_backend.py
denizdogan: Good catch, will fix.

denizdogan:
@acdha I've now updated the PR with your fixes applied!

acdha (approved):

acdha (approved):

mpflanzer:
Is there a specific reason that the `analyzer` keyword in only handled for the `TEXT` field type? Whoosh itself also allows it on the `KEYWORD` field type for example.

denizdogan:
@mpflanzer If the question is related to the PR, unfortunately I can't answer it. I don't even remember making this PR, it was so long ago!

PR #1793, opened 2021-06-10
Whoosh backend: join `AsyncWriter`s when they spawn threads

### LINKED ISSUE #1792: Updating whoosh index with --workers > 1 misses data
* [ ] Tested with the latest Haystack release

## Expected behaviour
Index item count should match the number of indexed objects

## Actual behaviour
Index item count is less than the number of indexed objects

## Description of actual problem

Using Whoosh as a backend, update_index:

https://github.com/django-haystack/django-haystack/blob/8ab14a273d8f438fe3542d988954e41bddc13fc8/haystack/management/commands/update_index.py#L119

calls into the Haystack Whoosh backend here:

https://github.com/django-haystack/django-haystack/blob/8ab14a273d8f438fe3542d988954e41bddc13fc8/haystack/backends/whoosh_backend.py#L247

which uses an AsyncWriter,

https://github.com/django-haystack/django-haystack/blob/8ab14a273d8f438fe3542d988954e41bddc13fc8/haystack/backends/whoosh_backend.py#L252

and later commits.

https://github.com/django-haystack/django-haystack/blob/8ab14a273d8f438fe3542d988954e41bddc13fc8/haystack/backends/whoosh_backend.py#L284-L286

For reference, here's whoosh.writing.AsyncWriter.commit:

https://github.com/mchaput/whoosh/blob/5421f1ab3bb802114105b3181b7ce4f44ad7d0bb/src/whoosh/writing.py#L1044-L1049

If the AsyncWriter managed to grab the lock on creation, commit commits and is done.

If the AsyncWriter did not manage to grab the lock on creation, a thread is spun up on `commit()` to try to grab the lock and flush out the writes. Nothing `join()`s that thread, so if the worker exits before its thread can make the writes, they are lost. There is likely also a race between the worker exiting and the thread writing.

I'll send a PR later, but it's going to be basically this:

https://github.com/django-haystack/django-haystack/blob/8ab14a273d8f438fe3542d988954e41bddc13fc8/haystack/backends/whoosh_backend.py#L284-L286

follow the `commit()` call with:
```python
if writer.ident is not None:
    writer.join()
```

Another plan might be to queue up all the threads and join at the end of the worker's tasks.

asedeno:
`whoosh.writing.AsyncWriter` has two behaviors, depending on whether
or not it succeeds at getting a write lock on the index when it is
initialized.

If it succeeds in getting the lock, it behaves as if it were
essentially `whoosh.writing.IndexWriter`.

If it fails to the get the lock, it buffers the writes, and on
`commit()` it spawns a thread to loop on trying to get the lock and
make the necessary writes.

In the latter case, a thread is left un`join()`ed, and may cause data
loss if the process exits before the thread can finish. This is
causing data loss in update_index with --workers > 1.

`woosh.writing.AsyncWriter` is also a vanilla python
`threading.Thread`, and we can tell if it has been started if
`writer.ident` is not None. If that's the case, we should `join()` it
to prevent writing a partial index.

Fixes #1792

### DIFF
--- a/haystack/backends/whoosh_backend.py
+++ b/haystack/backends/whoosh_backend.py
@@ -284,6 +284,8 @@ class WhooshSearchBackend(BaseSearchBackend):
         if len(iterable) > 0:
             # For now, commit no matter what, as we run into locking issues otherwise.
             writer.commit()
+            if writer.ident is not None:
+                writer.join()
 
     def remove(self, obj_or_string, commit=True):
         if not self.setup_complete:


asedeno:
This should also fix #860.

acdha (approved):

PR #1366, opened 2016-05-18
Basic Whoosh facet support

hellmanj:
resubmitting for @audax (original PR #914)  I didn't do anything except rebase to master

PR #1797, opened 2021-06-11
Whoosh faceting

asedeno:
Let's try this Whoosh faceting thing again. This picks up where #816 and #1366 left off. I've also worked in date faceting because I could.

I have not yet looked into Whoosh's `QueryFacet`.


asedeno:
@acdha, what can I do to help move faceting support for Whoosh in Haystack along?

acdha (approved):

Issue #1671, opened 2019-04-25
FacetedSearchView fails with elasticsearch 7.0.0


michaelmior:
* [*] Tested with the latest Haystack release
* [ ] Tested with the current Haystack master branch

## Expected behaviour

Search should work as it does with elasticsearch 6.3.1

## Actual behaviour

I get the exception below

```
Traceback (most recent call last):
  File "/home/travis/build/michaelmior/pylinks/pylinks/search/tests.py", line 20, in test_search
    response = self.client.get(reverse('haystack_search') + '?q=Google')
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/test/client.py", line 535, in get
    response = super().get(path, data=data, secure=secure, **extra)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/test/client.py", line 347, in get
    **extra,
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/test/client.py", line 422, in generic
    return self.request(**r)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/test/client.py", line 503, in request
    raise exc_value
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/python/3.6.3/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/haystack/views.py", line 51, in __call__
    return self.create_response()
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/haystack/views.py", line 150, in create_response
    context = self.get_context()
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/haystack/views.py", line 128, in get_context
    (paginator, page) = self.build_page()
[...]

PR #1800, opened 2021-06-16
Add a backend for ES7

Surgo:
Add a ES7 backend.

- Fixes: #1671


Surgo:
WIP

.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       matrix:
         django-version: [2.2, 3.1, 3.2]
         python-version: [3.6, 3.7, 3.8, 3.9]
-        elastic-version: [1.7, 2.4, 5.5]
+        elastic-version: [1.7, 2.4, 5.5, '7.13.1']
```
acdha, line 13: I'm wondering whether we want to pin this at a point version which presumably will change regularly or can just use `7.13`

acdha (approved):

PR #1825, opened 2021-11-30
document Elasticsearch 7.x compatibility

gregbaker:
Compatibility with Elasticsearch 7 was mentioned in the tutorial, but not a few other places in the docs.


claudep (approved):
Thanks!

PR #1834, opened 2022-02-02
Use standard Django test runner

claudep:


acdha (approved):

acdha:
Wait, I just realized that the time it was taking the tests to run was all setup — it looks like the run tests stage isn't finding anything so it exits immediately.

Issue #1650, opened 2018-11-07
can't update index with workers


fnsne:
* [ ] Tested with the latest Haystack release
* [ ] Tested with the current Haystack master branch

## Expected behaviour
If I use the command `python manage.py update_index--verbosity=2  --workers=3` , it should run mulitprocessing to update index.

## Actual behaviour
But, when I actual run `python manage.py update_index --verbosity=2 --workers=3`, I got error below.
```
[INFO/SpawnPoolWorker-3] child process calling self.run()
[INFO/SpawnPoolWorker-1] child process calling self.run()
[INFO/SpawnPoolWorker-2] child process calling self.run()
Process SpawnPoolWorker-3:
Process SpawnPoolWorker-1:
Traceback (most recent call last):
Traceback (most recent call last):
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\pool.py", line 108, in worker
    task = get()
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\pool.py", line 108, in worker
    task = get()
Process SpawnPoolWorker-2:
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\queues.py", line 337, in get
    return _ForkingPickler.loads(res)
  File "c:\users\watas\appdata\local\programs\python\python36-32\Lib\multiprocessing\queues.py", line 337, in get
    return _ForkingPickler.loads(res)
  File "C:\Users\watas\Documents\code\InteRed-2018\creators\models.py", line 6, in <module>
    class Channel(models.Model):
  File "C:\Users\watas\Documents\code\InteRed-2018\creators\models.py", line 6, in <module>
    class Channel(models.Model):
[...]

PR #1853, opened 2022-07-06
Proposal that resolves the issue issues #1650

alexsilva:
I've been having the same problem reported in #1650 and I made some fixes that solve the problem and still gives the possibility of other fixes through the configuration.


cclauss:
Please rebase now that we have more automated tests in place.

cclauss:
The tests seem to crash.

PR #1837, opened 2022-02-03
[WIP] Use standard Django test runner

acdha:
This is a follow to finish #1834 — I got overeager seeing the “WIP” cleared.

TODO:

* [ ] Remove wait-for-solr script from the other matrix builds
* [ ] Consider adding matrix for other Solr versions (6 is rather old by now)

PR #1855, opened 2022-09-18
GitHub Actions: Add Python 3.10 and 3.11 to the testing

cclauss:
'`3.10`' ___must___ be quoted in YAML.  https://dev.to/hugovk/the-python-3-1-problem-85g

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,25 +4,37 @@ on: [pull_request, push]
 
 jobs:
   test:
-
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
-      matrix:
+      fail-fast: false
+      matrix:  # https://docs.djangoproject.com/en/4.1/faq/install
         django-version: [2.2, 3.1, 3.2]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7]
         elastic-version: [1.7, 2.4, 5.5, '7.13.1']
         include:
           - django-version: '4.0'
-            python-version: 3.8
-            elastic-version: 5.5
+            python-version: '3.8'
+            elastic-version: '5.5'
           - django-version: '4.0'
-            python-version: 3.8
+            python-version: '3.8'
             elastic-version: '7.13.1'
           - django-version: '4.0'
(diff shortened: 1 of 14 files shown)


manelclos:
@cclauss tests are failing, can you please take a look?

cclauss:
There are two kinds of problems...
1. The flake8 errors look easy to fix.
2. The Py310 & Py311 errors look like they are caused by Nose not supporting new Python versions.  Two ways to go...
    1. [Nose2](https://pypi.org/project/nose2/) which is more difficult to move to than the name makes it sound.
    2. https://pypi.org/project/nose2pytest which is difficult to run but is probably a better long-term bet.

Pull requests to come...

claudep:
Note that #1837 is aiming at getting rid of `nose`. I guess that would be the PR to finish to solve recent Python failures.

cclauss:
Given https://docs.djangoproject.com/en/4.1/faq/install what do we ___really___ want to test?
Can we drop Django < 3.2?
Can we drop Python < 3.7?  Or < 3.8?
Can we drop Django-Haystack < 5.5?

cclauss:
# Blocked by #1837 or similar.

.github/workflows/test.yml
```diff
@@ -4,25 +4,37 @@ on: [pull_request, push]
...
-
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
-      matrix:
+      fail-fast: false
+      matrix:  # https://docs.djangoproject.com/en/4.1/faq/install
         django-version: [2.2, 3.1, 3.2]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7]
```
claudep, line 12: Really?

claudep: 3.6 is definitely outdated, let's remove. Then add 3.10 and 3.11. But maybe first address test failures with the current test matrix (without 3.6)?

claudep:
Do you intend to continue improving the patch to reduce failures, or do you think we should go ahead and fix failures in further PRs?

claudep:
FYI I approved and merged the pre-commit auto-update. You may rebase on latest master.

PR #1866, opened 2023-01-30
docs: use ‘stable’ tag in Django URLs | fix outdated URLs

HAMZA310:
In URLs that point to pages in django
documentation, use 'stable' tag in order to
always point to the latest version of the documentation. Previously, hard-coded django versions (e.g. 1.7)
were used in hyperlinks which are now outdated (throw 404) e.g. this link: https://docs.djangoproject.com/en/1.7/topics/class-based-views/ is no longer valid- used in [views_and_forms.html](https://django-haystack.readthedocs.io/en/master/views_and_forms.html)
page.

Using 'stable' tag will ensure those
links never become outdated.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -900,7 +900,7 @@ Other
   Add python 3.5 to tests
 - Add python 3.5 to tests. [Marco Badan]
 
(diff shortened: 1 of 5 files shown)


claudep (approved):
:+1:

PR #1871, opened 2023-04-07
Optimize recursive .all() query to get values only

mattegray:
### Issue
The recursive `current_objects.all()` is taking too long for fields with deep objects relations.

### Fix
Try to use `values_list` to get only the required values fast.

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -473,6 +473,58 @@ class MultiValueField(SearchField):
 
         return [value]
 
+    def resolve_attributes_lookup(self, current_objects, attributes):
+        """
+        Return the leaf node values for one or more iterable objects that contain data,
+        for model attributes that can be multiple objects (relations) deep.
+        """
+        values = []
+
+        for current_object in current_objects:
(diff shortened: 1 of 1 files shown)


haystack/fields.py
```diff
@@ -473,6 +473,58 @@ def convert(self, value):
...
+                current_object = getattr(current_object, attributes.pop(0))
+
+                if hasattr(current_object, "all") and ismethod(current_object.all):
+                    attribute = "__".join(attributes)
+                    return list(
+                        current_object.values_list(attribute, flat=True).distinct()
+                    )
+
+                elif not hasattr(current_object, "__iter__"):
+                    if hasattr(current_object, attributes[0]):
```
cclauss, line 499: Are there tests that come with these changes?  What will happen here if attributes is empty?

mattegray: @cclauss That code will run only when `len(attributes) > 1`, so I don't think there would be an issue?

cclauss: You are correct.  My mistake.

PR #1876, opened 2023-04-17
DRAFT: test_haystack/solr_tests/server/wait-for-solr

cclauss:
Let's try to the the tests passing again.


claudep:
AFAICS nothing is downloading/installing/running solr server in the current test setup. I guess that's what's missing.

cclauss:
I thought that GitHub Actions was starting a Docker-based service for Solr here...
https://github.com/django-haystack/django-haystack/blob/ffcd9bf8f20ca410e42d950114a99e4c557e726e/.github/workflows/test.yml#L47-L50

claudep:
Ah yes, good find. So I guess there is a communication issue between tests and the service...

cclauss:
I know little about Solr but that service is version 6 and the current is version 9.  I will experiment on a separate repo.

acdha:
Is it as simple as the Solr config using localhost:9001 when the service should be solr:9001 to get the service container? That script is old and presumably the default dates back to when we were running the JAR file for Solr directly.

acdha:
Thanks for looking into this, by the way!

test_haystack/solr_tests/server/wait-for-solr
```diff
@@ -8,7 +8,8 @@ import requests
 max_retries = 100
 retry_count = 0
 retry_delay = 15
-status_url = "http://localhost:9001/solr/collection1/admin/ping"
+# status_url = "http://localhost:9001/solr/collection1/admin/ping"
```
acdha, line 11: This will break anyone working locally. We should probably make this something like this:

```python
solr_url = os.environ.get("SOLR_URL", "http://localhost:9001/solr/collection1/")
status_url = f"{solr_url}admin/ping"
```

(we could do other things with the hostname but I figure this is easies to make it possible to customize the protocol, hostname, port, or collection name)

cclauss: I would not paint us into a corner.  I was just trying to see if your suggestion worked.  Unfortunately, it does not.
```
Unhandled exception requesting http://solr:9001/solr/collection1/admin/ping: HTTPConnectionPool(host='solr', port=9001): Max retries exceeded with url: /solr/collection1/admin/ping (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fbde5626a70>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
Waiting 15 seconds for Solr to start (retry #82, status 0)
```

acdha: I'm wondering how this originally worked because Solr's standard port is 8983 and we don't appear to be customizing that anywhere. I suspect that changing the port number on https://github.com/django-haystack/django-haystack/blob/ffcd9bf8f20ca410e42d950114a99e4c557e726e/.github/workflows/test.yml#L50 will fix that but that should have broken before now.

.github/workflows/test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       solr:
         image: solr:6
         ports:
-          - 9001:9001
+          - 8983:9001
```
claudep, line 50: Not sure, but could you try the reverse (9001:8983)?

cclauss, line 50:
```suggestion
          - 9001:8983
```

test_haystack/solr_tests/server/wait-for-solr
```diff
@@ -5,12 +5,15 @@ import time
...
 
-max_retries = 100
+max_retries = 10  # was 100
 retry_count = 0
 retry_delay = 15
 status_url = "http://localhost:9001/solr/collection1/admin/ping"
 
 
+# Fail fast if Solr is not up and running at port 9001
+assert requests.get("http://localhost:9001/solr").status_code == 200
```
cclauss, line 15: This line is passing so Solr is up and running at port 9001.

.github/workflows/test.yml
cclauss: Your suggestion was the correct answer.

cclauss:
% `docker ps`
````
CONTAINER ID   IMAGE                  COMMAND                  CREATED          STATUS                    PORTS                                                 NAMES
76eb4a1b6bd7   solr:9                 "docker-entrypoint.s…"   10 seconds ago   Up 9 seconds              0.0.0.0:8983->8983/tcp, :::8983->8983/tcp             test_solr
85d4b9a831b9   elasticsearch:7.17.9   "/bin/tini -- /usr/l…"   49 seconds ago   Up 48 seconds (healthy)   0.0.0.0:9200->9200/tcp, :::9200->9200/tcp, 9300/tcp   e76264a294684fe494dfe04d9afd8bda_elasticsearch7179_ed72ba
```

cclauss:
Closing in favor of
* #1960

PR #1896, opened 2023-08-28
Update spatial.rst

srivardhanrr:
# Fixed the import class for D, from django.contrib.gis.geos to django.contrib.gis.measure.

    # Previous Code to import "Point" and "D".
    from django.contrib.gis.geos import Point, D
    
    # to correct import paths of the packages.
    from django.contrib.gis.geos import Point
    from django.contrib.gis.measure import D

The correct package to import D is from django.contrib.gis.measure.


claudep (approved):
Thanks!

PR #1915, opened 2023-12-06
Django_v5.0

cclauss:
https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0
> The undocumented `django.utils.datetime_safe` module is removed.

Any projects that `import haystack` will fail without this change.

### DIFF
--- a/haystack/fields.py
+++ b/haystack/fields.py
@@ -2,7 +2,11 @@ import re
 from inspect import ismethod
 
 from django.template import loader
-from django.utils import datetime_safe
+
+try:  # datetime_safe was removed in Django 5.0
+    from django.utils import datetime_safe
+except ImportError:
+    import datetime as datetime_safe
 
 from haystack.exceptions import SearchFieldError
 from haystack.utils import get_model_ct_tuple


haystack/fields.py
```diff
@@ -2,7 +2,11 @@
 from inspect import ismethod
 
 from django.template import loader
-from django.utils import datetime_safe
+
+try:  # datetime_safe was removed in Django 5.0
+    from django.utils import datetime_safe
+except ImportError:
+    import datetime as datetime_safe
```
claudep, line 9: Looking at how Django made this replacement in its own code in https://github.com/django/django/commit/46346f8ea0802, it looks like it is always related to producing `strftime` ouptut, which is not the case in that file. So I would simply replace this by `import datetime` and replacing `datetime_safe` occurrences later in the file.

cclauss: It is always difficult to be confident about making changes to a project that is failing its CI.  I have taken a lot of runs at fixing this project's GitHub Actions but cannot seem to find the magic combination.

I have been using this branch to get all pytests of my main project to pass...
`Pipfile`:
```
# django-haystack = "*"  # Currently incompatible with Django 5.0
django-haystack = {git = "https://github.com/cclauss/django-haystack.git@Django_v5"}
```
but now that they are passing, let's try todo as you suggest above.

cclauss:
% `git grep datetime_safe`
```
haystack/backends/whoosh_backend.py:from django.utils.datetime_safe import date, datetime
test_haystack/whoosh_tests/test_whoosh_backend.py:from django.utils.datetime_safe import date, datetime
```
I did not change these instances because https://pypi.org/project/Whoosh has not been updated since 2016 and points to a nonexistent BitBucket repo however https://github.com/mchaput/whoosh still exists.

Issue #1924, opened 2023-12-18
Django 5 deprecated datetime_safe --> Drop support for whoosh?


itmve:
Using django 5 and django-haystack from git with the whoosh backend the following error occurs:

```
  File "/home/<project>/env/lib/python3.11/site-packages/haystack/backends/whoosh_backend.py", line 10, in <module>
    from django.utils.datetime_safe import date, datetime
ModuleNotFoundError: No module named 'django.utils.datetime_safe'
```
datetime_safe was deprectated in django 5.

cclauss:
https://pypi.org/project/Whoosh/ has not been updated since 2016 and points to a nonexistent repo on GitLab.

While https://github.com/mchaput/whoosh exists, it needs modernization.

There are lots of open issues in https://github.com/django-haystack/django-haystack/labels/backend%20%3A%20whoosh

cclauss:
Fixed by
* #1937

Issue #1917, opened 2023-12-07
Need Help: GitHub Actions tests are failing


cclauss:
# Fix container networking in our GitHub Actions environment.

Our GitHub Actions tests have consistently failed since they were set up.  Please help us to get effective tests that pass.
* https://github.com/django-haystack/django-haystack/actions/workflows/test.yml

This is caused by the issues in:
* #1921 
* #1929

Issue #149, opened 2009-12-05
Investigate issues with Whoosh & wildcard support


toastdriven:
From mailing list thread: http://groups.google.com/group/django-haystack/browse_frm/thread/300de8b7d3122392

Issue #23, opened 2009-05-06
Query fails if the backend decides everything is a stopword


apollo13:
http://paste.pocoo.org/show/115982/
I get this traceback searching for a single letter using whoosh as backend, the view should catch this, and tell me that my query is invalid. Apparently whoosh kicks out single letters, and as such my query is None

PR #1929, opened 2023-12-31
Drop support for whoosh

cclauss:
~Fixes: #1917~
Fixes: #1924
Fixes: #1928
* #1917
* #1924
* #1928

There are a lot of open issues under https://github.com/django-haystack/django-haystack/labels/backend%20%3A%20whoosh and the code has not been updated as discussed in #1924 so perhaps it is time to drop support for whoosh.
@nijel, @mchaput, @ZeroCool940711 Your reviews, please.

---

Fix assert calls for Python 3.12: https://docs.python.org/3/whatsnew/3.12.html#id3


ZeroCool940711:
I merged [PR #23](https://github.com/mchaput/whoosh/pull/23) from [mchaput/whoosh](https://github.com/mchaput/whoosh) repo in PR [#9](https://github.com/Sygil-Dev/whoosh/pull/9) of the [Sygil-Dev/whoosh](https://github.com/Sygil-Dev/whoosh) repo which fixed and issue boost in date/date queries, the problem on issue #1924 seems to be related more to haystack itself, still, Im not 100% sure tho. I do think that issue #149 might have been fixed with PR [#11](https://github.com/Sygil-Dev/whoosh/pull/11) tho since it adds support for wildcards.

claudep:
Should we open a discussion thread about the alternative between dropping whoosh or switching to whoosh-reloaded?

PR #1930, opened 2023-12-31
Fix unittest assert calls for Python 3.12

cclauss:
Fix `unittest` assert calls for Python 3.12: https://docs.python.org/3/whatsnew/3.12.html#id3
* Deprecated in Python 3.2 and removed in Python 3.12.

### DIFF
--- a/test_haystack/test_managers.py
+++ b/test_haystack/test_managers.py
@@ -242,11 +242,11 @@ class ManagerTestCase(TestCase):
 
     def test_values(self):
         sqs = self.search_index.objects.auto_query("test").values("id")
-        self.assertTrue(isinstance(sqs, ValuesSearchQuerySet))
+        self.assertIsInstance(sqs, ValuesSearchQuerySet)
 
     def test_valueslist(self):
         sqs = self.search_index.objects.auto_query("test").values_list("id")
-        self.assertTrue(isinstance(sqs, ValuesListSearchQuerySet))
+        self.assertIsInstance(sqs, ValuesListSearchQuerySet)
 
 
 class CustomManagerTestCase(TestCase):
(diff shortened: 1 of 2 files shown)


test_haystack/test_managers.py
```diff
@@ -242,11 +242,11 @@ def spelling_suggestion(self):
 
     def test_values(self):
         sqs = self.search_index.objects.auto_query("test").values("id")
-        self.assert_(isinstance(sqs, ValuesSearchQuerySet))
+        self.assertTrue(isinstance(sqs, ValuesSearchQuerySet))
```
claudep, line 245: We can use assertIsInstance instead, same below.

test_haystack/test_query.py
```diff
@@ -989,12 +989,12 @@ def test_valueslist_sqs(self):
...
         flat_sqs = self.msqs.auto_query("test").values_list("id", flat=True)
-        self.assert_(isinstance(sqs, ValuesListSearchQuerySet))
+        self.assertTrue(isinstance(sqs, ValuesListSearchQuerySet))
 
         # Note that this will actually be None because a mocked sqs lacks
         # anything else:
-        self.assert_(flat_sqs[0] is None)
-        self.assert_(flat_sqs[0:1][0] is None)
+        self.assertTrue(flat_sqs[0] is None)
+        self.assertTrue(flat_sqs[0:1][0] is None)
```
claudep, line 997: `assertIsNone` comes to mind.

claudep (changes requested):

test_haystack/test_managers.py
cclauss: ;-)  This is why most Python developers are delighted to drop all these different methods and allow pytest to take full advantage of assert rewriting.

cclauss:
@claudep Why disable auto-merge on this PR?  Is there something controversial about the proposed changes?

claudep (approved):

PR #1931, opened 2023-12-31
Update test matrix for django v5.0

cclauss:
https://pypi.org/project/Django

https://www.djangoproject.com/weblog/2023/dec/04/django-50-released

> Django 4.1 has reached the end of extended support. The final security release ([4.1.13](https://docs.djangoproject.com/en/stable/releases/4.1.13/)) was issued on November 1st. All Django 4.1 users are encouraged to [upgrade](https://docs.djangoproject.com/en/dev/howto/upgrade-version/) to Django 4.2 or later.

https://docs.djangoproject.com/en/5.0/releases/5.0
> Django 5.0 supports Python 3.10, 3.11, and 3.12.

https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django

### Whoosh issues with Python 3.10+ are fixed in #1928 or #1929
* #1928
* #1929

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     docs
-    py{38,39,310,311,312,py3}-django{3.2,4.2,5.0}-es{1.x,2.x,5.x,7.x}
+    py{38,39,310,311,312,py3}-django{3.2,4.2,5.0}-es7.x
 
 
 [testenv]
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     docs
-    py{36,37,38,39,310,py}-django{2.2,3.0,3.1,3.2,4.0}-es{1.x,2.x,5.x,7.x}
+    py{38,39,310,311,312,py3}-django{3.2,4.2,5.0}-es{1.x,2.x,5.x,7.x}
 
 
 [testenv]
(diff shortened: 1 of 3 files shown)


claudep:
It's nice to have a reference PR where tests pass, even if I'm reluctant to push as is. I still hope that we can make the solr tests pass at some point. May I ask you to make a new PR with only the version changes, that is mostly the changes in `test.yml`, `pyproject.toml` and `tox.ini`? Even it the tests are not green, we know now they are OK because of this one.

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     docs
-    py{36,37,38,39,310,py}-django{2.2,3.0,3.1,3.2,4.0}-es{1.x,2.x,5.x,7.x}
+    py{38,39,310,311,312,py3}-django{3.2,4.2,5.0}-es{1.x,2.x,5.x,7.x}
```
cclauss, line 4: @claudep Can we drop testing of `es{1.x,2.x,5.x}`

claudep: I think so.

cclauss:
```suggestion
    py{38,39,310,311,312,py3}-django{3.2,4.2,5.0}-es7.x
```

claudep (approved):
Thanks, let's go with this version that hopefully is a good step towards fixing the test suite!

cclauss:
> a good step towards fixing the test suite!

Fixing `wait-for-solr` and dropping `whoosh` should get us to ✅ 
* https://github.com/django-haystack/django-haystack/pull/1929/checks

PR #1510, opened 2017-05-23
Added logic to implement Interval Faceting (Solr only).

stmos:
This currently only works with Solr backend (>4.10).

More details about interval faceting can be found at: [Solr - Interval Faceting](https://cwiki.apache.org/confluence/display/solr/Faceting#Faceting-IntervalFaceting)

Example:
`intervals = ['[0,100]', '(100, *]']`
`sqs = SearchQuerySet()`
`sqs.interval_facet('price', intervals=intervals)`

Key replacement is supported as well:
` sqs.interval_facet('price', intervals=['{!key=cheap}[0, 100]', '{!key=expensive}[100, *)']).facet_counts()`

Output:
`{u'dates': {},
 u'fields': {},
 u'intervals': {'price': {u'cheap': 4, u'expensive': 47}},
 u'queries': {}}`

Syntax is explained at the URL posted above.
There is no need to specify the "facet=true"  when defining the search index for the fields you want to use for interval faceting.

Only requirement is that fields used for Interval Faceting must have “docValues” enabled (not required for Solr 5.20 and above)

PR #1934, opened 2024-01-07
Updated: Added logic to implement Interval Faceting (Solr only).

dwasyl:
This is just a re-based version of #1510 

Per stmos' original notes:
This currently only works with Solr backend (>4.10).

More details about interval faceting can be found at: [Solr - Interval Faceting](https://cwiki.apache.org/confluence/display/solr/Faceting#Faceting-IntervalFaceting)

Example:
intervals = ['[0,100]', '(100, *]']
sqs = SearchQuerySet()
sqs.interval_facet('price', intervals=intervals)

Key replacement is supported as well:
 sqs.interval_facet('price', intervals=['{!key=cheap}[0, 100]', '{!key=expensive}[100, *)']).facet_counts()

Output:
{u'dates': {}, u'fields': {}, u'intervals': {'price': {u'cheap': 4, u'expensive': 47}}, u'queries': {}}

Syntax is explained at the URL posted above.
There is no need to specify the "facet=true" when defining the search index for the fields you want to use for interval faceting.

Only requirement is that fields used for Interval Faceting must have “docValues” enabled (not required for Solr 5.20 and above)


dwasyl:
To deal with the test fail, is it worth refactoring the `build_search_kwargs` to use a generic `**kwargs`? I actually like it having as many clear arguments as it does, I find it makes the code very readable.

What about increasing the maximum number of allowed arguments in pylint?

cclauss:
> What about increasing the maximum number of allowed arguments in pylint?

Go ahead and increase `tool.ruff.pylint.max-args` in `pyproject.toml` so that reviewers can properly see all changes required.

test_haystack/test_managers.py
```diff
@@ -170,6 +170,11 @@ def test_date_facets(self):
         self.assertTrue(isinstance(sqs, SearchQuerySet))
         self.assertEqual(len(sqs.query.date_facets), 1)
 
+    def test_interval_facets(self):
+        sqs = self.search_index.objects.interval_facet("price", intervals="[0, 100]")
+        self.assertTrue(isinstance(sqs, SearchQuerySet))
```
cclauss, line 175: ;-)
```suggestion
        self.assertIsInstance(sqs, SearchQuerySet)
```

haystack/backends/solr_backend.py
```diff
@@ -296,6 +297,15 @@ def build_search_kwargs(
                     gap_by_string,
                 )
 
+        if interval_facets is not None:
+            kwargs["facet"] = "on"
+            kwargs["facet.interval"] = interval_facets.keys()
```
cclauss, line 302:
```suggestion
            kwargs["facet.interval"] = list(interval_facets)
```
Might be slightly safer...
```
>>> d = {key: value for value, key in enumerate("ABC")}
>>> d
{'A': 0, 'B': 1, 'C': 2}
>>> d.keys()[1]
  File "<stdin>", line 1, in <module>
TypeError: 'dict_keys' object is not subscriptable
>>> list(d)[1]
'B'
```

docs/searchqueryset_api.rst
```diff
@@ -323,6 +323,23 @@ Example::
...
     SearchQuerySet().filter(content='foo').facet('author')
 
+``interval_facet``
+~~~~~~~~~~~~~~
+
+.. method:: SearchQuerySet.interval_facet(self, field, interval)
+
+Adds interval faceting to a query for the provided field. You provide the field
+(from one of the ``SearchIndex`` classes) you like to facet on and an array of
+``intervals``. This is currently supported only when using Solr (ver 4.10+)
```
cclauss, line 333: Reviewers: Should we drop support for Solr < v4.10 or should this code assert that the Solr version is >= 4.10?

haystack/backends/solr_backend.py
dwasyl: Makes sense, although I'm not sure that it would ever be called that way (since you'd only use the `dict`), at the same time I don't know if there's a reason for it to a a `dict_keys` view either. I think technically `.keys()` is quicker since it's not making a new list, but in this case I can't see that ever being an issue. Happy to adjust, it seems to work either way on my running build.

claudep:
We first have to make the tests with solr pass again on master branch before being able to accept new functionalities related to solr.

dwasyl:
@claudep Totally understand, I was just updating my branch and figured it would be better to get this into the main release instead of being stuck on a fork. Hopefully you can working the solr testing issue, that's beyond me.

claudep:
> that's beyond me.

And beyond me, too :sob: Your intent to integrate your patch in the main branch is really appreciated, thanks!

dwasyl:
@claudep Is the testing updated and ready to consider adding this PR yet?

cclauss:
Please feel free to lift the simple lint change proposed in #2008 so all tests pass.
* #2008

dwasyl:
I rebased it, but the merge commits might have made the PR a mess.

If it's too much of a mess and it would be better to close and create a new one let me know.

dwasyl:
Just rebasing the branch.

PR #1935, opened 2024-01-12
Migrate away from pkg_resources

tacerus:
Using pkg_resources as an API is deprecated.
Migrate functionality to their importlib and packaging equivalents.

### DIFF
--- a/haystack/__init__.py
+++ b/haystack/__init__.py
@@ -1,7 +1,9 @@
+from importlib.metadata import PackageNotFoundError, version
+
 import django
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from pkg_resources import DistributionNotFound, get_distribution, parse_version
+from packaging.version import Version
 
 from haystack.constants import DEFAULT_ALIAS
 from haystack.utils import loading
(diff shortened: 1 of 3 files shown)


haystack/__init__.py
```diff
@@ -1,20 +1,21 @@
+from importlib.metadata import PackageNotFoundError, version
+
 import django
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from pkg_resources import DistributionNotFound, get_distribution, parse_version
+from packaging.version import Version
```
jayvdb, line 6: https://pypi.org/project/packaging/ is not listed as a dependency in `setup.py` .

There used to be a few Version classes in python's distutils, but that is removed in Python 3.12

Maybe add a dependency on `packaging` in setup.py ?

tacerus: Hi, thanks for the suggestion. I added it.

jayvdb (approved):

cclauss (approved):

cclauss:
This is needed because of the following error...
% `python3.12 -m pip install django-haystack`
% `python3.12`
>>> `import haystack`
```
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/haystack/__init__.py", line 4, in <module>
    from pkg_resources import DistributionNotFound, get_distribution, parse_version
ModuleNotFoundError: No module named 'pkg_resources'
```
https://docs.python.org/3/whatsnew/3.12.html#ensurepip

tacerus:
In our case it was needed to an update in `flufl-lock`, which broke several Python modules. Quite a story:
https://progress.opensuse.org/issues/153199

cclauss:
From the Py3.12 URL above I see that they recommend `pip install setuptools` instead of `pip install packaging` but for the build tools for Node.js add-ons we had to _vendor in_ `packaging` so I know that it works.
* nodejs/gyp-next#214

PR #1921, opened 2023-12-07
Set container URLs for tests running in GitHub Actions

acdha:
See #1917

PR #1938, opened 2024-01-22
Do not wait for solr

cclauss:
What happens after we solve:
* #1921


acdha:
This doesn't really solve the problem and the tests are still failing. Fixing the container networking in the actions environment should also resolve this issue with the wait-for-solr script.

PR #1943, opened 2024-02-01
Switch from Whoosh to Whoosh-Reloaded

cclauss:
# Blocked by 5 skipped tests plus disabled `wait_for_solr`...

https://pypi.org/project/Whoosh --> https://pypi.org/project/Whoosh-Reloaded
* https://github.com/Sygil-Dev/whoosh-reloaded/discussions/35

And stop waiting for Solr.
* #1921

~Blocked by Sygil-Dev/whoosh-reloaded#36~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
         django-version: ["3.2", "4.2", "5.0"]
-        python-version: ["3.8", "3.9"] # , "3.10", "3.11", "3.12"]  # Whoosh issues with Py3.10+
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"] # Whoosh issues with Py3.10+
         elastic-version: ["7.17.9"]
         exclude:
           - django-version: "3.2"
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ install_requires = [
 
 tests_require = [
     "pysolr>=3.7.0",
-    "whoosh>=2.5.4,<3.0",
+    "Whoosh-Reloaded>=2.7.5,<3.0",
     "python-dateutil",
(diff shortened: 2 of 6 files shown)


cclauss:
@ZeroCool940711 I could use your advice on this pull request vs. #1928.
* #1928

When I use the older [`Whoosh`](https://pypi.org/project/Whoosh) in #1928 then I only have __two failing tests to skip__.
However when I use [`Whoosh-Reloaded`](https://pypi.org/project/Whoosh-Reloaded) in this pull request then I have to skip those two plus an additional three for a total of __five failing tests to skip__.  Search, highlight, and dates are of particular concern.

Can you please look at the failing tests in this PR and #1928 to see if you have any advice on how to fix them?

Thanks much for all your work on `Whoosh-Reloaded` which does provide compatibility with Python >= 3.10.

ZeroCool940711:
> @ZeroCool940711 I could use your advice on this pull request vs. #1928.
> 
> * [Skip test_update_index_default_using() and test_rebuild_index_nocommit() #1928](https://github.com/django-haystack/django-haystack/pull/1928)
> 
> When I use the older [`Whoosh`](https://pypi.org/project/Whoosh) in #1928 then I only have **two failing tests to skip**. However when I use [`Whoosh-Reloaded`](https://pypi.org/project/Whoosh-Reloaded) in this pull request then I have to skip those two plus an additional three for a total of **five failing tests to skip**. Search, highlight, and dates are of particular concern.
> 
> Can you please look at the failing tests in this PR and #1928 to see if you have any advice on how to fix them?
> 
> Thanks much for all your work on `Whoosh-Reloaded` which does provide compatibility with Python >= 3.10.

I will look into it. I am trying to increase the test coverage as much as I can and I'm using several tools to do so as soon as possible, I'm also trying to reduce the code complexity, there are way too many loops and functions that are complex for pretty much no reason. For what I saw based on a few VS Code extensions I use for coverage and on Pytest itself we only have around 22% test coverage for the whole project, I think even that value might be too high for what we might actually have, the real value might be probably way lower than that.

PR #1953, opened 2024-03-28
Solr 9.5

seandhani:
I have a use case to support Solr 9.5.0 and noticed the the "app.model_name" string is being wrapped in a list. 
I've made the necessary updates, while maintaining compatibility while legacy versions of Solr.
I've also updated the Solr test server launcher to support the latest Solr releases.

Would appreciate some feedback.


seandhani:
Need to correct failing tests.

cclauss:
If you have any ideas on how to fix the failing tests, pull requests will be warmly received.

PR #1959, opened 2024-04-28
Fix SOLR setup under Github Actions

black3r:
See #1917 

As #1921 seems a bit stuck, I decided to pursue the same issue, but with smaller PRs. 

We don't really need container networking right now, because the tests run directly on the runner and not inside a container, and while it's nice to have, we can deal with that in a separate PR. Same goes for switching the SOLR port from the custom 9001 back to standard 8983.

This does just one thing - configures the SOLR properly under GH actions, so that the tests run. Also, unlike #1921, I did the configuration in a separate shell script, so that it can be also used for local testing with docker, for example with running solr with `docker run -it --rm -p 9001:8983 solr:6` and then running the same script locally...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install system dependencies
         run: sudo apt install --no-install-recommends -y gdal-bin
+      - name: Setup solr test server in Docker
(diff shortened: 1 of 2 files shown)


cclauss:
https://github.com/django-haystack/django-haystack/blob/ae7bd394b47a5102033b563e15577af58804203f/haystack/admin.py#L18

Change to `super().__init__(search_help_text="@cclauss, @black3r FIX ME!!!", **kwargs)`

black3r:
replaced by #1960

PR #1960, opened 2024-04-28
Fix failing tests

### LINKED ISSUE #1917: Need Help: GitHub Actions tests are failing

black3r:
fixes #1917 

this PR does the following:

- sets up Solr to work with Github Actions testing (using a setup script inspired by `start-solr-test-server.sh` and #1921) This setup script is also usable to run the test locally (for example with running solr with `docker run -it --rm -p 9001:8983 solr:6`)
- fixes `tox.ini` to properly configure `tox-gh-actions`
  - Tox in current master doesn't pick up the `DJANGO` env var set up by github actions, so it doesn't set up django according to `tox.ini` but instead just installs the `Django>=3.2` specified in `setup.py` ending up with all the tests running with Django 4.2, this fixes that... 
  - I removed `pypy3` from `tox.ini` because it seems unused currently. If it's actually needed, let me know and I will re-add it and add it to the github actions `test.yml` too... 
- fixes individual failing tests 
  - some tests were failing because haystack indexes weren't reset between the runs (= they would run okay individually but would error if they were ran after other test cases)..., maybe a better systemic solution would be to create a `HaystackTestCase` which would call the `.reset()` method on all indexes in its setUpClass method, similar to what django's `TestCase` does with databases, and subclass that for every testcase... for now I just fixed that in the affected tests... 
  - test for management command `build_solr_schema` used to depend on Solr server being set up locally, but for testing this command we don't actually need a live Solr server, we can mock the Solr server instead...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install system dependencies
         run: sudo apt install --no-install-recommends -y gdal-bin
+      - name: Setup solr test server in Docker
+        run: bash test_haystack/solr_tests/server/setup-solr-test-server-in-docker.sh
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip setuptools wheel
--- a/haystack/admin.py
+++ b/haystack/admin.py
@@ -18,7 +18,7 @@ class SearchChangeList(ChangeList):
         self.haystack_connection = kwargs.pop("haystack_connection", DEFAULT_ALIAS)
         super_kwargs = kwargs
         if django_version[0] >= 4:
-            super_kwargs['search_help_text'] = 'Search...'
+            super_kwargs["search_help_text"] = "Search..."
         super().__init__(**super_kwargs)
 
     def get_results(self, request):
--- a/haystack/admin.py
+++ b/haystack/admin.py
@@ -1,3 +1,4 @@
+from django import VERSION as django_version
 from django.contrib.admin.options import ModelAdmin, csrf_protect_m
 from django.contrib.admin.views.main import SEARCH_VAR, ChangeList
 from django.core.exceptions import PermissionDenied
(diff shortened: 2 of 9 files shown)


cclauss (approved):

claudep (commented):
Wow, that looks great! Now can you explain why the test runs shown by GitHub are only run for Python 3.8/3.9, Django 3.2/4.2 and ES?

cclauss:
> the test runs shown by GitHub are only run for Python 3.8/3.9, Django 3.2/4.2, and ES?

@black3r and I have agreed that we will make small moves so pull requests will be as small as possible to ease the code review process.

black3r:
> Wow, that looks great! Now can you explain why the test runs shown by GitHub are only run for Python 3.8/3.9, Django 3.2/4.2 and ES?

As the project is set up right now, the tests are started by GitHub actions and then handed over to Tox. So which tests are started is decided in `.github/actions/test.yml`. I didn't want to change this for now, but since there are more python and django versions defined in `tox.ini`, I prepared the `tox.ini` to accept all of them from GitHub Actions if we change that later...

DhavalGojiya:
CC: @black3r, @cclauss, @claudep, @acdha

Inspired by this PR, I’ve created my own **Setup Apache Solr GitHub Action** for GitHub Actions CI.

It sets up a complete Apache Solr infrastructure inside the GitHub runner, with automatic cleanup.  
It also supports custom configsets, custom core names, automatic health checks, and much more.
you just need to write **three lines** in your workflow!

If you’d like to take a look, here it is:  
👉 [setup-solr-action](https://github.com/DhavalGojiya/setup-solr-action)

PR #1962, opened 2024-04-29
Fixed whoosh test failures with Django 5.0

claudep:


test_haystack/whoosh_tests/test_whoosh_backend.py
```diff
@@ -114,6 +114,7 @@ def get_model(self):
         return MockModel
 
 
+@override_settings(USE_TZ=False)
```
cclauss, line 117: Gold star 🌟

cclauss (approved):

test_haystack/whoosh_tests/test_whoosh_backend.py
claudep: :smile: A git bisect was the key to found the culprit: https://github.com/django/django/commit/602d9a312facc64d

PR #1963, opened 2024-04-29
Add `setuptools` to fix failing tests on Python 3.12

cclauss:
Like #1962 but for Python 3.12
* #1962

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,7 @@ deps =
     geopy==2.0.0
     coverage
     requests
+    setuptools
     django3.2: Django>=3.2,<3.3
     django4.2: Django>=4.2,<4.3
     django5.0: Django>=5.0,<5.1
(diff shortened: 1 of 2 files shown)


claudep:
I guess a new pysolr release is needed to fix the issue on Python 3.12.

cclauss:
Just needed `setuptools` on Python 3.12.

tox.ini
```diff
@@ -28,6 +28,7 @@ deps =
     geopy==2.0.0
     coverage
     requests
+    setuptools
```
claudep, line 31: I would at least add a comment stating that the dependency could be dropped once pysolr 3.10 is released.

cclauss:
```suggestion
    setuptools; python_version >= "3.12"'  # Can be removed on pysolr >= v3.10
```

claudep:
Thanks, what about squashing commits?

cclauss:
I almost always `Squash and merge` unless the individual commits tell a story that is vital to preserve.

Screenshot 2024-04-30 at 11 31 59

claudep (approved):
Thanks!

PR #1982, opened 2024-07-09
pre-commit no longer supports the prettier file formater

cclauss:
https://github.com/pre-commit/mirrors-prettier is archived with the note:
> prettier made some changes that breaks plugins entirely

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,9 +46,3 @@ repos:
       - id: pretty-format-json
         args: ["--autofix", "--no-sort-keys", "--indent=4"]
       - id: trailing-whitespace
-
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
-    hooks:
-      - id: prettier
-        types_or: [json, toml, xml, yaml]


claudep (approved):
OK, for me. We are stuck by the indigestion of analysis tools :stuck_out_tongue:

acdha:
Yeah, that has been an unfortunate mess for a lot of projects and has me reconsidering the dependency of pre-commit. On some projects, I've worked around this with this local hook config:

```yaml
  - repo: local
    hooks:
      - id: prettier
        name: prettier
        language: node
        entry: npx --yes prettier@latest --write --ignore-unknown
        types_or: [css, markdown, javascript, jsx, ts, tsx]
        pass_filenames: true
```

However, I don't love that since since it's less stable. Looking at Haystack currently, we have some JSON files which can be handled entirely by the built-in JSON formatter which only needs Python, a single `pyproject.toml` file, and two YAML files:

```
.readthedocs.yaml
.pre-commit-config.yaml
```

That makes me wonder whether we should use a GitHub action to format those and otherwise not spend time on a custom hook.

cclauss:
> it's less stable

If we replaced `prettier@latest` with `prettier@3.3.2` then it should be fairly stable.  Or am I missing something?

Issue #1986, opened 2024-07-16
Build Solr schema : LatLonType has been removed in Solr 9


HugoDel:
* [X] Tested with the latest Haystack release
* [ ] Tested with the current Haystack master branch

## Expected behaviour
Configure a fixed scheme using `build_solr_schema` command and load it into Solr to create a core

## Actual behaviour
An error occur when creating the core using the `scheme.xml` auto-generated file.
`solr.LatLonType` used in the schema file but it's deprecated since Solr 7 and remove in Solr 9. [See Solr release note](https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-7.html#spatial-fields). It should be replaced with [LatLonPointSpatialField](https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-7.html#spatial-fields)

## Steps to reproduce the behaviour

1. Create a django model with a GeoDjango Point Field
2. Run the `build_solr_schema` command
3. Start a Solr core with the output file

## Configuration

* Operating system version: docker image solr:9.6.1-slim (Debian jammy)
* Search engine version: Solr 9.6.1
* Python version: 3.10.12
* Django version: 4.2.12
* Haystack version: 3.3.0

andresmrm:
I also had problems trying to use Solr 9.7 with django-haystack current config templates. They were related with these types/features:
- schema.xml:
	- LatLong
	- location
	- currency.xml
- solrconfig.xml
	- XSLT
	- elevate.xml

The configs worked after removing these parts. I posted the resulting files here:
https://gist.github.com/andresmrm/041f2724ded216c10b7d809fa7ea8a4e

Then, I wanted to make the solrconfig the closest possible to the default Solr 9.7 config, to use the newest features. Also made the schema the smallest possible (removing most types) and forced Portuguese lang. Resulting files here:
https://gist.github.com/andresmrm/20dc66395d1e0a0fd911863aadd4c950

It should also be possible to use the schema from the first link (EN and with most types) with the solrconfig from the second link (Solr 9.7 defaults).

You can use these files placing them in `<your app>/templates/search_configuration/`

Disclaimer: I have almost zero Solr experience. :upside_down_face: So be careful when using these files.

cclauss:
Our GitHub Actions do not run against modern versions of Solr.  If you could help us to fix 
* #1976
* #1983 
* #1984

then we would be able to properly validate fixes against modern Solr.

[...]

PR #2000, opened 2024-11-01
An attempt to fix the config error in new Solr versions

andresmrm:
Related to: https://github.com/django-haystack/django-haystack/issues/1986#issuecomment-2452016110

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,9 +20,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
-        django-version: ["3.2", "4.2", "5.0"]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        django-version: ["5.0"]  # ["3.2", "4.2", "5.0"]  # TODO Undo
+        python-version: ["3.12"]  # ["3.8", "3.9", "3.10", "3.11", "3.12"]  # TODO Undo
         elastic-version: ["7.17.9"]
+        solr-version: [6, 7, 8, 9]
         exclude:
           - django-version: "3.2"
             python-version: "3.11"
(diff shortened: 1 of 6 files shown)


test_haystack/solr_tests/test_solr_backend.py
```diff
@@ -1639,11 +1639,11 @@ def test_boost(self):
         )
 
         self.assertEqual(
-            [result.id for result in results],
+            sorted([result.id for result in results]),
```
cclauss, line 1642:
```suggestion
            sorted(result.id for result in results),
```
This is one of the ruff rules in https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

andresmrm: Thanks!

andresmrm:
The last (I hope) problem seems to be with the extraction endpoint:
```
ERROR: test_content_extraction (test_haystack.solr_tests.test_solr_backend.LiveSolrContentExtractionTestCase)
----------------------------------------------------------------------
  File "/home/runner/work/django-haystack/django-haystack/test_haystack/solr_tests/test_solr_backend.py", line 1670, in test_content_extraction
    self.assertTrue("haystack" in data["contents"])
TypeError: argument of type 'NoneType' is not iterable
```

andresmrm:
I'm not sure, but I think this is a problem in pysolr. I'm guessing it expects the content in a field `<filename>`, but this changed to `file`. https://github.com/django-haystack/pysolr/blob/master/pysolr.py#L1287
The metadata field structure also seems to have changed.
I could handle these changes inside the test code, but, if what I said is true, it will probably continue to cause errors in other places.
I'll see if I can boot a Solr 7 to check.

andresmrm:
I think the fix for pysolr would be something like this: https://github.com/django-haystack/pysolr/pull/490

andresmrm:
Hi, @cclauss ! Any update on this? :smiley:

andresmrm:
Hi, @acdha ! Would you have time to take a look at this? :grin:

PR #2008, opened 2025-04-27
Fix ruff rule C420 unnecessary-dict-comprehension-for-iterable

cclauss:
https://docs.astral.sh/ruff/rules/unnecessary-dict-comprehension-for-iterable

## Why is this bad?
It's unnecessary to use a dict comprehension to build a dictionary from an iterable when the value is static.

Prefer `dict.fromkeys(iterable)` over `{value: None for value in iterable}`, as `dict.fromkeys` is more readable and efficient.


cclauss:
Closing as fixed in #2007.
* #2007

PR #2025, opened 2025-09-04
Dev: Speeds up test by over a minute once the cache has been populated

kingbuzzman:
Installing `gdal-bin` is extremely slow, best case scenario [34s](https://github.com/django-haystack/django-haystack/actions/runs/17387285912/job/49355756978#step:5:1), average around [1:30s-ish](https://github.com/django-haystack/django-haystack/actions/runs/17387285912/job/49355756982#step:5:1), and i've seen it go as high as 4m+... [cant find it now]. I propose we use this action, the best case scenario remains in the 30s

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,7 +95,7 @@ jobs:
         run: |
           echo "$(pwd)/.gdal-root/usr/bin" >> $GITHUB_PATH
           {
-            echo "LD_LIBRARY_PATH=$(pwd)/.gdal-root/usr/lib/x86_64-linux-gnu:$(pwd)/.gdal-root/usr/lib:$(pwd)/.gdal-root/lib/x86_64-linux-gnu:${LD_LIBRARY_PATH}"
+            echo "LD_LIBRARY_PATH=$(pwd)/.gdal-root/usr/lib/x86_64-linux-gnu:$(pwd)/.gdal-root/usr/lib:$(pwd)/.gdal-root/lib/x86_64-linux-gnu:$(pwd)/.gdal-root/usr/lib/x86_64-linux-gnu/blas:$(pwd)/.gdal-root/usr/lib/x86_64-linux-gnu/lapack:${LD_LIBRARY_PATH}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -53,9 +54,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install system dependencies
-        run: |
-          sudo apt update
-          sudo apt install --no-install-recommends -y gdal-bin
+        uses: daaku/gh-action-apt-install@49782cd9703eaa5c436b9e6caa6529e7c5e50ecd  # v4
```
acdha, line 57: I'm not a big fan of linking to individual user's GitHub Actions with the whole supply-chain fun going around the open source world (not that I have any reason to suspect this guy has been compromised). It looks like this could be replaced with a single call to the official `actions/cache` action:

```yaml
    - id: cache
      uses: actions/cache@v4
      with:
        path: ~/.apt-cache
        key: apt-cache-${{ runner.os }}
```

cclauss: Agreed.  This is an exposure to an uncertain software layer.

cclauss:
> This branch is out-of-date with the base branch

.github/workflows/test.yml
```diff
@@ -52,10 +53,57 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install system dependencies
+      - name: Restore GDAL runtime cache
+        id: gdal-cache
+        uses: actions/cache@v4
+        with:
+          path: .gdal-root
+          key: gdal-${{ runner.os }}-jammy-gdalbin-v2
+      - name: Build portable GDAL (cache miss)
```
acdha, line 62: This seems like it quickly grew a lot more complex – why do we need to do more than install `gdal-bin` normally?

kingbuzzman: This is still a draft 😉  -- trying things out, it takes too long for the tests to run currently 😇 

Can you look at my open PRs please? 😇

PR #2031, opened 2025-10-21
Add Python 3.14 and 3.14t to the testing

cclauss:
https://www.python.org/downloads/release/python-3140/
https://py-free-threading.github.io/porting

### DIFF
--- a/test_haystack/run_tests.py
+++ b/test_haystack/run_tests.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+import multiprocessing
 import os
 import sys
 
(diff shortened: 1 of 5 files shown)


test_haystack/run_tests.py
```diff
@@ -11,6 +12,11 @@ def run_all(argv=None):
     os.environ["DJANGO_SETTINGS_MODULE"] = "test_haystack.settings"
     django.setup()
 
+    # Python >= 3.14: default start method changed from `fork` --> `forkserver`
```
acdha: Django 6 is supposed to handle this for the test runner: https://code.djangoproject.com/ticket/36531

That makes me wonder whether the answer for this is something like either not testing older versions of Django against the 3.14 free-threaded builds or possibly having a runtime warning when loading Haystack with Django 5 when `sysconfig.get_config_var("Py_GIL_DISABLED")` until we're comfortable that update_index doesn't have any subtle edge cases with free-threading.

cclauss: Python 3.14 is the first version that has a supported free-threaded build (`python3.14t`).

Python < 3.13 had no free-threaded build.  Python 3.13 had an unsupported free-threaded build (`python3.13t`).

acdha: I guess my question is whether our test matrix should include Django 5 on Python 3.14, or say 3.14 requires 6 to match the upstream.

acdha (approved):

acdha (approved):

PR #2040, opened 2026-01-01
chore: Ignore generated haystack version file

DhavalGojiya:
### Description

This PR updates .gitignore to exclude `haystack/version.py`, which is auto-generated by setuptools-scm during build/install.


DhavalGojiya:
> Are these lines problematic?

Yes, they are hard-coded at the moment and may lead to inconsistencies.

PR #2041, opened 2026-01-01
ci: Remove ruff job from the test ci pipeline

DhavalGojiya:
## Description

This PR removes the Ruff job from the test CI pipeline.

Ruff checks are already handled by [pre-commit.ci](https://pre-commit.ci/), so running them in the test pipeline was redundant and unnecessarily increased CI runtime.


cclauss:
Both ruff runs are sub-second, and one runs on a pinned version of ruff while the other runs on the current ruff.  This means that they may discover different issues, which provides useful information.

DhavalGojiya:
> Both ruff runs are sub-second, and one runs on a pinned version of ruff while the other runs on the current ruff. This means that they may discover different issues, which provides useful information.

Understood, Thanks. @cclauss

PR #2047, opened 2026-02-05
delegate building auto_query to backend.

fgregg:
Related to https://github.com/django-haystack/django-haystack/discussions/2046, here's a PR that delegates the auto_query building to the backend.

It's backwards compatible, but the tests also provide a motivating example for when we would want more control then the current design of AutoQuery.prepare provides.

### DIFF
--- a/haystack/backends/__init__.py
+++ b/haystack/backends/__init__.py
@@ -823,7 +823,7 @@ class BaseSearchQuery:
             elif rough_token not in exacts:
                 # We have something that's not an exact match but may have more
                 # than one word in it.
-                tokens.extend(rough_token.split(" "))
+                tokens.extend(rough_token.split())
             else:
                 tokens.append(rough_token)
 
--- a/haystack/backends/__init__.py
+++ b/haystack/backends/__init__.py
@@ -1,4 +1,5 @@
 import copy
+import re
 from copy import deepcopy
 from time import time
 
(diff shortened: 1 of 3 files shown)


haystack/backends/__init__.py
```diff
@@ -798,6 +799,47 @@ def build_not_query(self, query_string):
...
+        exacts = exact_match_re.findall(query_string)
+        tokens = []
+
+        for rough_token in exact_match_re.split(query_string):
+            if not rough_token:
+                continue
+            elif rough_token not in exacts:
+                # We have something that's not an exact match but may have more
+                # than one word in it.
+                tokens.extend(rough_token.split(" "))
```
cclauss:
```suggestion
                tokens.extend(rough_token.split())
```
Safer with returns, line feeds, tabs, multiple spaces, etc.
```
>>> "  a  b  c  d\nn\rr\tt  ".split()
['a', 'b', 'c', 'd', 'n', 'r', 't']
>>> "  a  b  c  d\nn\rr\tt  ".split(" ")
['', '', 'a', '', 'b', '', 'c', '', 'd\nn\rr\tt', '', '']
```

fgregg: good idea. i was just aiming at straight backwards compatibility, but might as well improve the behavior while we have the books open.

PR #2048, opened 2026-02-10
Fix race condition in ConnectionRouter.routers lazy initialization

craigds:
Build the router list in a local variable before assigning to `self._routers`, so concurrent threads never see a partially-initialized empty list.

Previously, `self._routers = []` was set before routers were appended. If another thread accessed the `routers` property between the empty-list assignment and the loop completing, it would see `_routers` as `[]` (not `None`), skip initialization, and return no routers. This caused `SearchIndex.get_backend()` to return `None`

### DIFF
--- a/haystack/utils/loading.py
+++ b/haystack/utils/loading.py
@@ -143,10 +143,11 @@ class ConnectionRouter:
             if not len(router_list):
                 router_list = default_routers
 
-            self._routers = []
+            routers = []
             for router_path in router_list:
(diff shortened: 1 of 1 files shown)


acdha (approved):

acdha (approved):
This should fix the error. I was wondering whether we wanted to add some kind of lock to prevent multiple threads running the setup code but it shouldn't be the case that the router initialization is expensive enough to warrant that for something which happens only at the start of the process.

PR #2059, opened 2026-06-04
publish a wheel, as well as an sdist

dimbleby:
Publish a wheel


acdha (approved):
This makes sense – wheels are going to install faster even for a pure-Python project.

PR #2068, opened 2026-08-14
eol: Drop support for end-of-life Python 3.9 and 3.10

DhavalGojiya:
## Description

_Drop support for end-of-life Python **3.9** and **3.10**_

See: https://endoflife.date/python

Python 3.10 is only 2 months away from reaching EOL.

image


cclauss:
The current recommended changes are to add Django 6.1 and Python 3.14 and drop Python 3.9.

I would be reluctant to drop support for Python 3.10 before we add support for Python 3.15.

DhavalGojiya:
> The current recommended changes are to add Django 6.1 and Python 3.14 and drop Python 3.9.

Okay, for now I will scope this PR to only dropping **Python 3.9**
I'll add support for Django 6.1 in the next PR.

PR #2069, opened 2026-08-14
eol: Drop support for end-of-life Python 3.9

DhavalGojiya:
## Description

_Drop support for end-of-life Python **3.9**_

See: https://endoflife.date/python

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
         django-version: ["4.2", "5.1", "5.2", "6.0"]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         elastic-version: ["7.17.12"]
         include:
           - django-version: "5.2"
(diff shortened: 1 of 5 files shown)


cclauss (approved):
Nice!  In a future pull request we should  upgrade elasticv 7.17 and add a comment which explains the licensing reasons why we do not use elastic v8 or v9.

DhavalGojiya:
> Nice! In a future pull request we should upgrade elasticv 7.17 and add a comment which explains the licensing reasons why we do not use elastic v8 or v9.

I check the ElasticSearch license drama starting from Es8, does this mean django-haystack will continue to support only Elasticsearch 7.x and won't add support for Elasticsearch 8 or 9 in the future?

cclauss:
I would be OK with a `single include test run` that tested the latest Elastic v9 and the latest Django 6.1 on the latest Python 3.14t.  However, I believe the rest of our testing should remain focused on the last and final version of Open Source Elastic v7.17.

DhavalGojiya:
> Open Source Elastic v7.17

Okay, but I don't think many people use the EOL Elasticsearch 7 version anymore. For newer Elasticsearch versions such as 8 and 9, people already use the official Elasticsearch Python client and django-elasticsearch-dsl.

From what I have seen, this django-haystack package has been unmaintained for the past 2 to 3 years and is now mostly used with Apache Solr.

I can help add support for newer Apache Solr versions once we release the new pysolr version, which supports both Solr 9 and Solr 10.

cclauss:
The project that I originally came to `django-haystack` for still uses the Open Source Elastic v7.

DhavalGojiya:
> The project that I originally came to `django-haystack` for still uses the Open Source Elastic v7.

That's great, but the numbers are very low.

I am using django-haystack with Solr in several 8+ projects. Two of them are US government projects.

Btw, I will create a PR to test the latest patch version of Elasticsearch 7.17 in CI.

cclauss:
I am not surprised that the US government has purchased the software.  Not everyone has the same resources.

acdha:
> I am not surprised that the US government has purchased the software. Not everyone has the same resources.

Even within the U.S. government, not everyone has those resources…

